PR #3, opened 2020-11-01
Adds support for "fyne release"

### LINKED ISSUE #1: Add support for "fyne release"
Before 1.0.0 we should support the "release" version of packages.
This partly is internal detail ("fyne release" instead of "fyne package") but also will require some new flags to be passed in for certificates / passwords.

This is still a work in progress upstream, but should be ready in a week or two.

lucor:
This PR adds support for "fyne release"

- [x] Update fyne cli to v1.4.2-0.20201127180716-f9f91c194737 fyne-io/fyne#1609 fyne-io/fyne#1527 fyne-io/fyne#1538 (Release 20.11.28)
- [x] Add support to: AppBuild, Release and AppVersion flags
- [x] Add support to "fyne release" for android os
- [x] Add support to "fyne release" for windows os
- [x] Add support to "fyne release" for ios os
- [x] Add support to "fyne release" for macOS os

> Note: since "release" for windows, ios and macOS requires native tools, it is supported only on the relative hosts via the fyne-cli tool.

Fixes #1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 - Add support for "fyne release" #3
 - Add support for creating packaged .tar.gz bundles on freebsd #6
-- Update fyne cli to v1.4.2-0.20201122132119-67b762f56dc0 fyne-io#1527
+- Update fyne cli to v1.4.2-0.20201125075943-97ad77d2abe0 (fyne-io#1538 fyne-io#1527)
 
 ## [0.9.0]
 - Releaseing under project namespace with previous 2.2.1 becoming 0.9.0 in fyne-io namespace
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog - Fyne.io fyne-cross
 
+## Unreleased
+- Add support for "fyne release" #3
+- Add support for creating packaged .tar.gz bundles on freebsd #6
+- Update fyne cli to v1.4.2-0.20201122132119-67b762f56dc0 fyne-io#1527
+
 ## [0.9.0]
 - Releaseing under project namespace with previous 2.2.1 becoming 0.9.0 in fyne-io namespace
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,9 @@ Supported targets are:
   -  android
   -  ios
 
-> Note: iOS compilation is supported only on darwin hosts. See [fyne README mobile](https://github.com/fyne-io/fyne/blob/v1.2.4/README-mobile.md#ios) for pre-requisites.
+> Note: 
+> - iOS compilation is supported only on darwin hosts. See [fyne pre-requisites](https://developer.fyne.io/started/#prerequisites) for details.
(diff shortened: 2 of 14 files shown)


andydotxyz (commented):
Great work getting this together thanks.
I see that iOS and Windows release build is called on the host, so I guess this means that they can lookup the keystore / key file for the signing.
However the Android build asks for a keystore path - is that accessible from the runtime ? (It may be outside the project) - and it will ask for the password on the command line, does that work in this model?

I was a little confused as fyneRelease vs fyneReleaseHost seems to have duplicated code but as I understand it the former is just android? In which location do linux/bsd releases run?

lucor:
> Great work getting this together thanks.
> I see that iOS and Windows release build is called on the host, so I guess this means that they can lookup the keystore / key file for the signing.

Right, it is basically a wrapper for `fyne release` on the host for Windows and iOS. This is needed since the tools used signign are not available into the docker image.

> However the Android build asks for a keystore path - is that accessible from the runtime ? (It may be outside the project)

Yes, but it can only access files inside the project. Mounting of external files or directories is not supported yet.

> - and it will ask for the password on the command line, does that work in this model?

Good catch, no won't work. We need to specify the password with a flag like for Windows. If this is not supported by `fyne release` we could always allow the Andoid release only in host mode for now.

> I was a little confused as fyneRelease vs fyneReleaseHost seems to have duplicated code but as I understand it the former is just android? In which location do linux/bsd releases run?

Yep, I was unsure about the naming too. Tempted to add the `Docker` suffix but at the end I kept fyneRelease for consintence with the fynePackage. Anyway renaming to fyneReleaseDocker and fynePackageDocker is an easy fix if it helps to improve the code readability.

TD;DR
fynePackage and fyneRelease run the `fyne` tool inside the container.
fynePackageHost and fyneReleaseHost run the `fyne` tool from the host.
Only Windows (release command) and iOS (package and release) are executed on the host.

andydotxyz:
Testing locally, did a --pull, but I don't think the bundled fyne is new enough?

```
❯ fyne-cross darwin -release
[i] Target: darwin/amd64
[✓] "bin" dir cleaned: /Users/andy/Go/src/github.com/fyne-io/life/fyne-cross/bin/darwin-amd64
[✓] "dist" dir cleaned: /Users/andy/Go/src/github.com/fyne-io/life/fyne-cross/dist/darwin-amd64
[✓] "temp" dir cleaned: /Users/andy/Go/src/github.com/fyne-io/life/fyne-cross/tmp/darwin-amd64
[i] Checking for go.mod: /Users/andy/Go/src/github.com/fyne-io/life/go.mod
[✓] go.mod found
[i] Packaging app...
flag provided but not defined: -category
Usage: fyne [command] [parameters], where command is one of:
...
```

andydotxyz (approved):
👍

andydotxyz:
> flag provided but not defined: -category
> Usage: fyne [command] [parameters], where command is one of:

For the record this was an out of date local fyne, which is used by darwin release.

PR #22, opened 2020-12-26
Fix release flags behavior

lucor:
- Fix android keystore path is not resolved correctly
- Fix some release flags are always set even if empty


stuartmscott (approved):
Thanks for the fix!

andydotxyz (approved):

PR #23, opened 2020-12-29
Fix listed arches for linux build

Jacalz:
It was listing the supported arches for Windows and thus did not list arm and arm64.


lucor (approved):

Jacalz:
FYI, @lucor I have no push access in this repo. You will have to merge this ;)

PR #36, opened 2021-03-13
Fix the option --env does not allow values containing comma

### LINKED ISSUE #35: the option --env prevent the environment GOPROXY from being set to more than one proxy
`GOPROXY` supports more than one proxy, which are separated by the comma. But the option `--env` also requires that the arguments must be separated by the comma. They are not compatible with each other.

```shell
$ fyne-cross windows -env GOPROXY='https://goproxy.io,direct'
invalid value "GOPROXY=https://goproxy.io,direct" for flag -env: env var must defined as KEY=VALUE or KEY=
```

After removing `,direct`, it works.

**Question:** How will I set the environment `GOPROXY` to many proxies by the option `--env` or others.

lucor:
Fixes #35


Jacalz (commented):
How are you supposed to set multiple environment values from the terminal after this change?

lucor:
> How are you supposed to set multiple environment values from the terminal after this change?

It is possible to set multiple env vars using one flag per var, like the docker cli command:
```
fyne-cross linux --env KEY1=value1 --env KEY2=value2 
```

andydotxyz (approved):
Yes, I think this is the right way forward

PR #42, opened 2021-04-22
Propose the ability to use podman

metal3d:
- fix the ldflags to be set in GOFLAGS, so there is no problem by
escaping environment variables
- ldflags "-H" should be set with equal sign `-H=windowsgui` instead of
`-H windowsgui'
- add podman detection
- add podman usage: no need to fix uid with gosu, podman has got an
option named "--userns" to set to "keep-id".

### DIFF
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -3,7 +3,9 @@ package command
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"os/exec"
 	"os/user"
 	"runtime"
 	"strings"
@@ -13,6 +15,9 @@ import (
 	"github.com/fyne-io/fyne-cross/internal/volume"
 	"golang.org/x/sys/execabs"
 )
+
+// IsPodman is set by init() function, and is true if podman is used
+var IsPodman bool
 
 const (
 	// fyneBin is the path of the fyne binary into the docker image
@@ -30,6 +35,11 @@ func CheckRequirements() error {
 	return nil
 }
 
+func init() {
+	// Initialise podman detection once
+	IsPodman = initIsPodman()
+}
+
 // Options define the options for the docker run command
 type Options struct {
 	CacheEnabled bool     // CacheEnabled if true enable go build cache
@@ -38,15 +48,28 @@ type Options struct {
 	Debug        bool     // Debug if true enable log verbosity
 }
 
-// IsPodman returns true if docker is an alias to podman (e.g. via podman-docker).
-func IsPodman() bool {
-	out, err := exec.Command("docker", "info", "-f", "{{.Host.RemoteSocket.Path}}").Output()
+// initIsPodman returns true if docker is an alias to podman (e.g. via podman-docker).
+func initIsPodman() bool {
+	// read the "docker" executable to check if it's a podman wrapper
+
+	execPath, err := exec.LookPath("docker")
+	if err != nil {
+		// docker is not installed, is there "podman" ?
+		if _, err = exec.LookPath("podman"); err == nil {
+			return true
+		}
+	}
+
+	// OK, let's see if "docker" comes from "podman-docker" aliases
+	f, err := os.Open(execPath)
 	if err != nil {
 		return false
 	}
-	// There should be an indication about the remote socket that is commonly named
-	// like /path/to/podman.sock
-	return strings.Contains(string(out), "podman")
+	defer f.Close()
+
+	// if it's a script alias, so "docker" is a bash script that calls "podman"
+	c, _ := ioutil.ReadAll(f)
+	return strings.Contains(string(c), "podman")
 }
 
 // Cmd returns a command to run in a new container for the specified image
@@ -64,8 +87,7 @@ func Cmd(image string, vol volume.Volume, opts Options, cmdArgs []string) *execa
 		"-v", fmt.Sprintf("%s:%s:z", vol.WorkDirHost(), vol.WorkDirContainer()), // mount the working dir
 	}
 
-	if IsPodman() {
-		log.Info("USE PODMAN")
+	if IsPodman {
 		args = append(args, "--userns", "keep-id", "-e", "use_podman=1")
 	}
 
@@ -101,7 +123,11 @@ func Cmd(image string, vol volume.Volume, opts Options, cmdArgs []string) *execa
 	args = append(args, cmdArgs...)
 
 	// run the command inside the container
-	cmd := execabs.Command("docker", args...)
+	c := "docker"
+	if IsPodman {
+		c = "podman"
+	}
+	cmd := exec.Command(c, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
@@ -165,7 +191,20 @@ func goBuild(ctx Context) error {
 	if len(ldflags) > 0 {
 		flags := make([]string, len(ldflags))
 		for i, flag := range ldflags {
-			flags[i] = "-ldflags=" + flag
+			if strings.Index(flag, "-X") == 0 {
+				// We must rebuild cases
+				// if the user pass "-ldflags "-X A.B=C" so we need to set it to "-X=A.B=C"
+				// if the user pass "-ldflags "-X=A.B=C" so we should not change it
+				sp := strings.SplitN(flag, " ", 1)
+				if len(sp) == 2 {
+					args = append(args, "-ldflags=-X="+sp[1])
+				} else {
+					args = append(args, "-ldflags="+flag)
+				}
+			} else {
+				// others flags can be set to GOFLAGS
+				flags[i] = "-ldflags=" + flag
+			}
 		}
 
 		// ensure that GOFLAGS is not overwritten as they can be passed
@@ -178,6 +217,7 @@ func goBuild(ctx Context) error {
 			ctx.Env = append(ctx.Env, fmt.Sprintf("GOFLAGS=%s", strings.Join(flags, " ")))
 		}
 	}
+
 	// add tags to command, if any
 	tags := ctx.Tags
 	if len(tags) > 0 {
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -46,7 +46,7 @@ func IsPodman() bool {
 	}
 	// There should be an indication about the remote socket that is commonly named
 	// like /path/to/podman.sock
-	return strings.Index(string(out), "podman") > -1
+	return strings.Contains(string(out), "podman")
 }
 
 // Cmd returns a command to run in a new container for the specified image
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -202,19 +202,22 @@ func goBuild(ctx Context) error {
 	if len(ldflags) > 0 {
 		flags := make([]string, len(ldflags))
 		for i, flag := range ldflags {
-			if strings.HasPrefix(flag, "-X") {
-				// We must rebuild cases
-				// if the user pass "-ldflags "-X A.B=C" so we need to set it to "-X=A.B=C"
-				// if the user pass "-ldflags "-X=A.B=C" so we should not change it
-				sp := strings.SplitN(flag, " ", 2)
-				if len(sp) == 2 {
-					args = append(args, "-ldflags=-X="+sp[1])
+			parts := strings.Split(flag, "-") // because there could be several -X flags
+			for _, p := range parts {
+				if strings.HasPrefix(p, "X") {
+					// We must rebuild cases
+					// if the user pass "-ldflags "-X A.B=C" so we need to set it to "-X=A.B=C"
+					// if the user pass "-ldflags "-X=A.B=C" so we should not change it
+					sp := strings.SplitN(p, " ", 1)
+					if len(sp) == 2 {
+						args = append(args, "-ldflags=-X="+sp[1])
+					} else {
+						args = append(args, "-ldflags=-"+p)
+					}
 				} else {
-					args = append(args, "-ldflags="+flag)
+					// others flags can be set to GOFLAGS
+					flags[i] = "-ldflags=-" + p
 				}
-			} else {
-				// others flags can be set to GOFLAGS
-				flags[i] = "-ldflags=" + flag
 			}
 		}
 
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -2,7 +2,6 @@ package command
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"os"
 	"os/user"
@@ -140,14 +139,14 @@ func goModInit(ctx Context) error {
 	return nil
 }
 
-// findGoFlags return the index of context where GOFLAGS is set.
-func findGoFlags(ctx Context) (int, error) {
+// findGoFlags return the index of context where GOFLAGS is set, or -1 if it is not present.
+func findGoFlags(ctx Context) int {
 	for i, env := range ctx.Env {
-		if strings.Index(env, "GOFLAGS") == 0 {
-			return i, nil
+		if strings.HasPrefix(env, "GOFLAGS") {
+			return i
 		}
 	}
-	return -1, errors.New("GOFLAGS not found in context")
+	return -1
 }
 
 // goBuild run the go build command in the container
@@ -171,7 +170,7 @@ func goBuild(ctx Context) error {
 
 		// ensure that GOFLAGS is not overwritten as they can be passed
 		// by he --env argument
-		if idx, err := findGoFlags(ctx); err == nil {
+		if idx := findGoFlags(ctx); idx > -1 {
 			// append the ldflags after the existing GOFLAGS
 			ctx.Env[idx] += " " + strings.Join(flags, " ")
 		} else {
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -191,7 +191,7 @@ func goBuild(ctx Context) error {
 	if len(ldflags) > 0 {
 		flags := make([]string, len(ldflags))
 		for i, flag := range ldflags {
-			if strings.Index(flag, "-X") == 0 {
+			if strings.HasPrefix(flag, "-X") {
 				// We must rebuild cases
 				// if the user pass "-ldflags "-X A.B=C" so we need to set it to "-X=A.B=C"
 				// if the user pass "-ldflags "-X=A.B=C" so we should not change it
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -14,13 +15,6 @@ import (
 	"github.com/fyne-io/fyne-cross/internal/log"
 	"github.com/fyne-io/fyne-cross/internal/volume"
 	"golang.org/x/sys/execabs"
-)
-
-var (
-	// IsPodman is set by init() function, and is true if podman is used
-	IsPodman bool
-	// Runner is either "docker" or "podman"
-	Runner string
 )
 
 const (
@@ -45,16 +39,6 @@ func CheckRequirements() error {
 	return nil
 }
 
-func init() {
-	// Initialise podman detection once
-	IsPodman = initIsPodman()
-	if IsPodman {
-		Runner = "podman"
-	} else {
-		Runner = "docker"
-	}
-}
-
 // Options define the options for the docker run command
 type Options struct {
 	CacheEnabled bool     // CacheEnabled if true enable go build cache
@@ -63,28 +47,42 @@ type Options struct {
 	Debug        bool     // Debug if true enable log verbosity
 }
 
-// initIsPodman returns true if docker is an alias to podman (e.g. via podman-docker).
-func initIsPodman() bool {
-	// read the "docker" executable to check if it's a podman wrapper
+// engine returns the default engine name, if no engine is found it returns an empty string and the error.
+func engine() (string, error) {
+	if isEngineDocker() {
+		return "docker", nil
+	}
 
-	execPath, err := exec.LookPath("docker")
-	if err != nil {
-		// docker is not installed, is there "podman" ?
-		if _, err = exec.LookPath("podman"); err == nil {
-			return true
+	if isEnginePodman() {
+		return "podman", nil
+	}
+
+	return "", errors.New("No container engine found")
+
+}
+
+// isEnginePodman returns true if the engine is "podman"
+func isEnginePodman() bool {
+	_, err := exec.LookPath("podman")
+	return err == nil
+
+}
+
+// isEngineDocker return true is the engine is "docker". If "docker" is an alias to podman, so it returns false.
+func isEngineDocker() bool {
+	if execPath, err := exec.LookPath("docker"); err == nil {
+		// OK, let's see if "docker" comes from "podman-docker" aliases
+		f, err := os.Open(execPath)
+		if err != nil {
+			return false
 		}
-	}
+		defer f.Close()
 
-	// OK, let's see if "docker" comes from "podman-docker" aliases
-	f, err := os.Open(execPath)
-	if err != nil {
-		return false
+		// if it's a script alias, so "docker" is a bash script that calls "podman"
+		c, _ := ioutil.ReadAll(f)
+		return !strings.Contains(string(c), "podman")
 	}
-	defer f.Close()
-
-	// if it's a script alias, so "docker" is a bash script that calls "podman"
-	c, _ := ioutil.ReadAll(f)
-	return strings.Contains(string(c), "podman")
+	return false
 }
 
 // Cmd returns a command to run in a new container for the specified image
@@ -102,7 +100,7 @@ func Cmd(image string, vol volume.Volume, opts Options, cmdArgs []string) *execa
 		"-v", fmt.Sprintf("%s:%s:z", vol.WorkDirHost(), vol.WorkDirContainer()), // mount the working dir
 	}
 
-	if IsPodman {
+	if isEnginePodman() {
 		args = append(args, "--userns", "keep-id", "-e", "use_podman=1")
 	}
 
@@ -138,7 +136,11 @@ func Cmd(image string, vol volume.Volume, opts Options, cmdArgs []string) *execa
 	args = append(args, cmdArgs...)
 
 	// run the command inside the container
-	cmd := exec.Command(Runner, args...)
+	runner, err := engine()
+	if err != nil {
+		log.Fatal(err.Error())
+	}
+	cmd := exec.Command(runner, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
@@ -452,7 +454,11 @@ func pullImage(ctx Context) error {
 	buf := bytes.Buffer{}
 
 	// run the command inside the container
-	cmd := exec.Command(Runner, "pull", registry+"/"+ctx.DockerImage)
+	runner, err := engine()
+	if err != err {
+		log.Fatal(err.Error())
+	}
+	cmd := exec.Command(runner, "pull", registry+"/"+ctx.DockerImage)
 	cmd.Stdout = &buf
 	cmd.Stderr = &buf
 
@@ -460,7 +466,7 @@ func pullImage(ctx Context) error {
 		log.Debug(cmd)
 	}
 
-	err := cmd.Run()
+	err = cmd.Run()
 
 	if ctx.Debug {
 		log.Debug(buf.String())
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -16,14 +16,20 @@ import (
 	"golang.org/x/sys/execabs"
 )
 
-// IsPodman is set by init() function, and is true if podman is used
-var IsPodman bool
+var (
+	// IsPodman is set by init() function, and is true if podman is used
+	IsPodman bool
+	// Runner is either "docker" or "podman"
+	Runner string
+)
 
 const (
 	// fyneBin is the path of the fyne binary into the docker image
 	fyneBin = "/usr/local/bin/fyne"
 	// gowindresBin is the path of the gowindres binary into the docker image
 	gowindresBin = "/usr/local/bin/gowindres"
+	// registry is the docker registry to use to pull images
+	registry = "docker.io"
 )
 
 // CheckRequirements checks if the docker binary is in PATH
@@ -42,6 +48,11 @@ func CheckRequirements() error {
 func init() {
 	// Initialise podman detection once
 	IsPodman = initIsPodman()
+	if IsPodman {
+		Runner = "podman"
+	} else {
+		Runner = "docker"
+	}
 }
 
 // Options define the options for the docker run command
@@ -121,17 +132,13 @@ func Cmd(image string, vol volume.Volume, opts Options, cmdArgs []string) *execa
 	}
 
 	// specify the image to use
-	args = append(args, image)
+	args = append(args, registry+"/"+image)
 
 	// add the command to execute
 	args = append(args, cmdArgs...)
 
 	// run the command inside the container
-	c := "docker"
-	if IsPodman {
-		c = "podman"
-	}
-	cmd := exec.Command(c, args...)
+	cmd := exec.Command(Runner, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
@@ -442,7 +449,7 @@ func pullImage(ctx Context) error {
 	buf := bytes.Buffer{}
 
 	// run the command inside the container
-	cmd := exec.Command("docker", "pull", "docker.io/"+ctx.DockerImage)
+	cmd := exec.Command(Runner, "pull", registry+"/"+ctx.DockerImage)
 	cmd.Stdout = &buf
 	cmd.Stderr = &buf
 
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -438,7 +438,7 @@ func pullImage(ctx Context) error {
 	buf := bytes.Buffer{}
 
 	// run the command inside the container
-	cmd := execabs.Command("docker", "pull", ctx.DockerImage)
+	cmd := exec.Command("docker", "pull", "docker.io/"+ctx.DockerImage)
 	cmd.Stdout = &buf
 	cmd.Stderr = &buf
 
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -38,6 +38,17 @@ type Options struct {
 	Debug        bool     // Debug if true enable log verbosity
 }
 
+// IsPodman returns true if docker is an alias to podman (e.g. via podman-docker).
+func IsPodman() bool {
+	out, err := exec.Command("docker", "info", "-f", "{{.Host.RemoteSocket.Path}}").Output()
+	if err != nil {
+		return false
+	}
+	// There should be an indication about the remote socket that is commonly named
+	// like /path/to/podman.sock
+	return strings.Index(string(out), "podman") > -1
+}
+
 // Cmd returns a command to run in a new container for the specified image
 func Cmd(image string, vol volume.Volume, opts Options, cmdArgs []string) *execabs.Cmd {
 
@@ -50,12 +61,17 @@ func Cmd(image string, vol volume.Volume, opts Options, cmdArgs []string) *execa
 	args := []string{
 		"run", "--rm", "-t",
 		"-w", w, // set workdir
-		"-v", fmt.Sprintf("%s:%s", vol.WorkDirHost(), vol.WorkDirContainer()), // mount the working dir
+		"-v", fmt.Sprintf("%s:%s:z", vol.WorkDirHost(), vol.WorkDirContainer()), // mount the working dir
+	}
+
+	if IsPodman() {
+		log.Info("USE PODMAN")
+		args = append(args, "--userns", "keep-id", "-e", "use_podman=1")
 	}
 
 	// mount the cache dir if cache is enabled
 	if opts.CacheEnabled {
-		args = append(args, "-v", fmt.Sprintf("%s:%s", vol.CacheDirHost(), vol.CacheDirContainer()))
+		args = append(args, "-v", fmt.Sprintf("%s:%s:z", vol.CacheDirHost(), vol.CacheDirContainer()))
 	}
 
 	// add default env variables
@@ -137,9 +153,12 @@ func goBuild(ctx Context) error {
 
 	// add ldflags to command, if any
 	if len(ldflags) > 0 {
-		args = append(args, "-ldflags", fmt.Sprintf("'%s'", strings.Join(ldflags, " ")))
+		flags := make([]string, len(ldflags))
+		for i, flag := range ldflags {
+			flags[i] = fmt.Sprintf("-ldflags=%s", flag)
+		}
+		ctx.Env = append(ctx.Env, fmt.Sprintf("GOFLAGS=%s", strings.Join(flags, " ")))
 	}
-
 	// add tags to command, if any
 	tags := ctx.Tags
 	if len(tags) > 0 {
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -28,9 +28,13 @@ const (
 
 // CheckRequirements checks if the docker binary is in PATH
 func CheckRequirements() error {
-	_, err := execabs.LookPath("docker")
+	_, err := exec.LookPath("docker")
+	if err == nil {
+		return nil
+	}
+	_, err = exec.LookPath("podman")
 	if err != nil {
-		return fmt.Errorf("missed requirement: docker binary not found in PATH")
+		return fmt.Errorf("missed requirement: docker or podman binary not found in PATH")
 	}
 	return nil
 }
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"os/user"
@@ -139,6 +140,16 @@ func goModInit(ctx Context) error {
 	return nil
 }
 
+// findGoFlags return the index of context where GOFLAGS is set.
+func findGoFlags(ctx Context) (int, error) {
+	for i, env := range ctx.Env {
+		if strings.Index(env, "GOFLAGS") == 0 {
+			return i, nil
+		}
+	}
+	return -1, errors.New("GOFLAGS not found in context")
+}
+
 // goBuild run the go build command in the container
 func goBuild(ctx Context) error {
 	log.Infof("[i] Building binary...")
@@ -157,7 +168,16 @@ func goBuild(ctx Context) error {
 		for i, flag := range ldflags {
 			flags[i] = fmt.Sprintf("-ldflags=%s", flag)
 		}
-		ctx.Env = append(ctx.Env, fmt.Sprintf("GOFLAGS=%s", strings.Join(flags, " ")))
+
+		// ensure that GOFLAGS is not overwritten as they can be passed
+		// by he --env argument
+		if idx, err := findGoFlags(ctx); err == nil {
+			// append the ldflags after the existing GOFLAGS
+			ctx.Env[idx] += " " + strings.Join(flags, " ")
+		} else {
+			// create the GOFLAGS environment variable
+			ctx.Env = append(ctx.Env, fmt.Sprintf("GOFLAGS=%s", strings.Join(flags, " ")))
+		}
 	}
 	// add tags to command, if any
 	tags := ctx.Tags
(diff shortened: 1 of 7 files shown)


lucor:
Thanks for the PR it looks good to me, just few thoughts.

PRs  should be have `develop` as base branch, could you please switch to it ?

The GOFLAGS could be also set via the fyne-cross `--env` flag, so we should ensure that internal GOFLAGS value are always appended to the ones specified by the user, if any. ([rightmost setting wins](https://github.com/golang/go/issues/38522#issuecomment-716212148))

metal3d:
I changed the base, so there is a conflict I will resolve.

I will check for GOFLAG.

metal3d:
Sorry there were a non fixed conflict, I will resolve (and fix the test)

metal3d:
@lucor I made a new test to fix GOFLAGS if it is sent in parameters. I can see:
```
$ fyne-cross linux  --debug
/usr/bin/docker run --rm -t -w /app -v /tmp/foo:/app:z -v /home/pafer/.cache/fyne-cross:/go:z -e CGO_ENABLED=1 \
-e GOCACHE=/go/go-build \
-e GOOS=linux -e GOARCH=amd64 -e CC=gcc \
-e GOFLAGS=-ldflags=-w -ldflags=-s \
-e fyne_uid=1000 fyneio/fyne-cross:1.1-base go build -o /app/fyne-cross/bin/linux-amd64/foo -v .
```
And with `--env` added:
```
$ fyne-cross linux --env GOFLAGS="-ldflags=-X='main.Version=1.0.0'" --debug
/usr/bin/docker run --rm -t -w /app -v /tmp/foo:/app:z -v /home/pafer/.cache/fyne-cross:/go:z -e CGO_ENABLED=1 \
-e GOCACHE=/go/go-build \
-e GOFLAGS=-ldflags=-X='main.Version=1.0.0' -ldflags=-w -ldflags=-s \
-e GOOS=linux -e GOARCH=amd64 -e CC=gcc \
-e fyne_uid=1000 fyneio/fyne-cross:1.1-base go build -o /app/fyne-cross/bin/linux-amd64/foo -v .
```

lucor (commented):
I've tried the changes using a Version as per your example. 
I was able to see it into the logs, but in the fyne app that uses that variable seems to be ignored :(
Wondering if the multiple -ldflags are allowed.
Additionally I forgot that there is a fyne-cross `--ldflags` that we should handle as well.

internal/command/docker.go
```diff
@@ -123,6 +140,16 @@ func goModInit(ctx Context) error {
...
 
+// findGoFlags return the index of context where GOFLAGS is set.
+func findGoFlags(ctx Context) (int, error) {
+	for i, env := range ctx.Env {
+		if strings.Index(env, "GOFLAGS") == 0 {
+			return i, nil
+		}
+	}
+	return -1, errors.New("GOFLAGS not found in context")
+}
```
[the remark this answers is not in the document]
Jacalz: Also, in the same file. Couldn’t we just use strings.Contains?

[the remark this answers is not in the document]
metal3d: @lucor I did it the "go" way, but yes, it's possible to only return the index that is -1

@Jacalz because I want to be sure that this is the environment variable that is found and nothing more. It is possible that you want to pass "GOFLAGS" as a string in `-X` options (for example to inject a message in a variable at build time to say "with GOFLAGS")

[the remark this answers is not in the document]
metal3d: @lucor I don't understand your first question. I was able to set multiple `-ldflags` - to set `Version` I also made the test and that worked with: `-e GOFLAGS=-ldflags=-X='main.Version=1.0.0'`

internal/command/docker.go
```diff
@@ -137,9 +163,21 @@ func goBuild(ctx Context) error {
...
+		flags := make([]string, len(ldflags))
+		for i, flag := range ldflags {
+			flags[i] = fmt.Sprintf("-ldflags=%s", flag)
+		}
 
+		// ensure that GOFLAGS is not overwritten as they can be passed
+		// by he --env argument
+		if idx := findGoFlags(ctx); idx > -1 {
+			// append the ldflags after the existing GOFLAGS
+			ctx.Env[idx] += " " + strings.Join(flags, " ")
```
Jacalz, line 175: Seems like a combination of two different ways. I would suggest either doing `" " + flags + " "` or perhaps even `strings.Join(" ", flags, " ")`.

Jacalz (commented):
Looking forward to seeing this landed. Just a few comments/suggestions for you.

internal/command/docker.go
Jacalz: Ah, I thought that `strings.Join()` did something different. Ignore this one.

internal/command/docker.go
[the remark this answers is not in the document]
lucor: > I did it the "go" way, but yes, it's possible to only return the index that is -1 if not found

Thanks for the changes. I see and agree with you about the Go way to return the error. In that specific case I'm ok with a little exception in favour of saving an err check like the `strings.Index` method due the fact it won't return error.

> don't understand your first question. I was able to set multiple -ldflags - to set Version I also made the test and that worked with: -e GOFLAGS=-ldflags=-X='main.Version=1.0.0'

I tested locally with a small fyne app and tried to "inject" to variable values with ldflags but despite the fact I was able to see the flags in the fyne-cross compilation output, it was not set for the app.

Code I've used to test:
```
cat main.go 
package main

import (
	"fyne.io/fyne/v2"
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/container"
	"fyne.io/fyne/v2/widget"
)

var Title = "Default"
var Name = "Default"

func main() {
	a := app.New()
	w := a.NewWindow(Title)

	hello := widget.NewLabel("Hi " + Name)
	w.SetContent(container.NewVBox(
		hello,
	))
	w.Resize(fyne.NewSize(200, 200))

	w.ShowAndRun()
}
```

Command used to compile:
```
fyne-cross linux --env GOFLAGS="-ldflags=-X='main.Name=Podman'" --debug .
```

fyne-cross logs:
```
[i] Building binary...
/usr/bin/docker run --rm -t -w /app -v /code/lucor/test:/app:z -v /home/lucor/.cache/fyne-cross:/go:z -e CGO_ENABLED=1 -e GOCACHE=/go/go-build -e GOFLAGS=-ldflags=-X='main.Name=Podman' -ldflags=-w -ldflags=-s -e GOOS=linux -e GOARCH=amd64 -e CC=gcc -e fyne_uid=1000 fyneio/fyne-cross:1.1-base go build -o /app/fyne-cross/bin/linux-amd64/test -v .
[✓] Binary: /code/lucor/test/fyne-cross/bin/linux-amd64/test
[i] Packaging app...
/usr/bin/docker run --rm -t -w /app -v /code/lucor/test:/app:z -e CGO_ENABLED=1 -e GOCACHE=/go/go-build -e fyne_uid=1000 fyneio/fyne-cross:1.1-base /usr/local/bin/fyne version
fyne cli version v2.0.2
/usr/bin/docker run --rm -t -w /app/fyne-cross/tmp/linux-amd64 -v /code/lucor/test:/app:z -v /home/lucor/.cache/fyne-cross:/go:z -e CGO_ENABLED=1 -e GOCACHE=/go/go-build -e GOFLAGS=-ldflags=-X='main.Name=Podman' -ldflags=-w -ldflags=-s -e GOOS=linux -e GOARCH=amd64 -e CC=gcc -e fyne_uid=1000 fyneio/fyne-cross:1.1-base /usr/local/bin/fyne package -os linux -name test -icon /app/fyne-cross/tmp/linux-amd64/Icon.png -appBuild 1 -appVersion 1.0 -executable /app/fyne-cross/bin/linux-amd64/test
[✓] Package: /code/lucor/test/fyne-cross/dist/linux-amd64/test.tar.gz
```

Application output:
image

Please also notes, that fyne-cross exposes also a `-ldflags` so we should have the following command to work too:

```fyne-cross linux -ldflags="-X=main.Name=Podman" --debug .```


Wrt the multilple ldflags, it seems that when you specify them like:
```
GOFLAGS="-ldflags=-X='main.Name=Podman' -ldflags=-s -ldflags=-w" go build
```
or 
```
go build -ldflags=-X='main.Name=Podman' -ldflags=-s -ldflags=-w
```
only the last flag is used (try to compile changing the order)

The following instead work as expected:

```
go build -ldflags "-X=main.Name=Podman -w -s"
```

**Review by Jacalz — CHANGES_REQUESTED** [review]
Thanks again for working on this @metal3d. It is a very nice feature to have and would be nice to see it landed soon. Do you think you could get this rebased on develop? The tests don't want to start and I think the out-of-date branch might partly be the reason.

I finally managed to test this out on my new Fedora 34 installation. I had to install `podman-docker` to even get it started. Would it be possible to get it to only use `podman`? After that, I ran into this error:
```
~/go/bin/fyne-cross linux -arch amd64,arm64 -app-id com.github.jacalz.wormhole-gui -icon internal/assets/icon/icon-512.png
[i] Target: linux/amd64
[✓] "bin" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/bin/linux-amd64
[✓] "dist" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/dist/linux-amd64
[✓] "temp" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/tmp/linux-amd64
[i] Checking for go.mod: /home/jacob/Git/wormhole-gui/go.mod
[✓] go.mod found
[i] Building binary...
USE PODMAN
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Error: error getting default registries to try: short-name resolution enforced but cannot prompt without a TTY
[✗] exit status 125
make: *** [Makefile:51: linux] Error 1
```

**Comment by metal3d** [comment]
I will soon. Sorry for the delay, very busy these days...

File: internal/command/docker.go [hunk 3]
```diff
@@ -137,9 +163,21 @@ func goBuild(ctx Context) error {
 
 	// add ldflags to command, if any
 	if len(ldflags) > 0 {
-		args = append(args, "-ldflags", fmt.Sprintf("'%s'", strings.Join(ldflags, " ")))
-	}
+		flags := make([]string, len(ldflags))
+		for i, flag := range ldflags {
+			flags[i] = fmt.Sprintf("-ldflags=%s", flag)
```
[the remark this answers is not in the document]
metal3d: Done in https://github.com/fyne-io/fyne-cross/pull/42/commits/dc404256d85f2922a8aa45011a091479585f1631 ;)

internal/command/docker.go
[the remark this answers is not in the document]
metal3d: > Thanks again for working on this @metal3d. It is a very nice feature to have and would be nice to see it landed soon. Do you think you could get this rebased on develop? The tests don't want to start and I think the out-of-date branch might partly be the reason.
> 

I'm currently makeing the merges and conflct resolution. I hope to not make regression...

> I finally managed to test this out on my new Fedora 34 installation. I had to install `podman-docker` to even get it started. Would it be possible to get it to only use `podman`? After that, I ran into this error:
> 

This is what I discussed earlier - I wanted to use the "podman" command instead of "docker" with "podman-docker" mackage. Yes, of course, the best is to use the "podman" command IMHO. I can make some changes for this.

> ```
> ~/go/bin/fyne-cross linux -arch amd64,arm64 -app-id com.github.jacalz.wormhole-gui -icon internal/assets/icon/icon-512.png
> [i] Target: linux/amd64
> [i] Cleaning target directories...
> [✓] "bin" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/bin/linux-amd64
> [✓] "dist" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/dist/linux-amd64
> [✓] "temp" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/tmp/linux-amd64
> [i] Checking for go.mod: /home/jacob/Git/wormhole-gui/go.mod
> [✓] go.mod found
> [i] Building binary...
> USE PODMAN
> Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
> Error: error getting default registries to try: short-name resolution enforced but cannot prompt without a TTY
> [✗] exit status 125
> make: *** [Makefile:51: linux] Error 1
> ```

It's been a while I didn't make tests with my fyne-cross changes, I don't remember how I tested 😅 

I will try to make some tests a soon as the confict and merge / rebase are OK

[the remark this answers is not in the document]
metal3d: OK, I see the problem.

This is the test file:
```go
package main

import (
	"fyne.io/fyne/v2/app"
	"fyne.io/fyne/v2/widget"
)

var Message = "Default"

func main() {
	app := app.New()
	w := app.NewWindow(Message)
	t := widget.NewLabel(Message)

	w.SetContent(t)
	w.ShowAndRun()
}
```

When I build with:
```
../../fyne-cross/fyne-cross linux -ldflags="-X main.Message=42"
```

The error is:
```
go: parsing $GOFLAGS: non-flag "main.Message=42"
```


If I do:
```
# note the "=" after -X
../../fyne-cross/fyne-cross linux -ldflags="-X=main.Message=42"
```

So it builds, but the `main.Message` is not set...

it seems to be related to https://github.com/golang/go/issues/26849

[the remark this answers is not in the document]
metal3d: OK... I have new information.

The documentation says:

> GOFLAGS
>	A space-separated list of -flag=value settings to apply
>	to go commands by default, when the given flag is known by
>	the current command. Each entry must be a standalone flag.
>	Because the entries are space-separated, flag values must
>	not contain spaces. Flags listed on the command line
>	are applied after this list and therefore override it.

So, "flags value" cannot contain spaces... And `-ldflag=-X main.Value=foo` is badly interpreted. This is the only one ldflag that will not work with `-X=main...`

That means that  this particular ldflag shoud be passed in the command line, not in environment. 

So, what I can try is to pass ldflags to the command line...

metal3d:
> Thanks again for working on this @metal3d. It is a very nice feature to have and would be nice to see it landed soon. Do you think you could get this rebased on develop? The tests don't want to start and I think the out-of-date branch might partly be the reason.
> 
> I finally managed to test this out on my new Fedora 34 installation. I had to install `podman-docker` to even get it started. Would it be possible to get it to only use `podman`? After that, I ran into this error:
> 
> ```
> ~/go/bin/fyne-cross linux -arch amd64,arm64 -app-id com.github.jacalz.wormhole-gui -icon internal/assets/icon/icon-512.png
> [i] Target: linux/amd64
> [i] Cleaning target directories...
> [✓] "bin" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/bin/linux-amd64
> [✓] "dist" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/dist/linux-amd64
> [✓] "temp" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/tmp/linux-amd64
> [i] Checking for go.mod: /home/jacob/Git/wormhole-gui/go.mod
> [✓] go.mod found
> [i] Building binary...
> USE PODMAN
> Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
> Error: error getting default registries to try: short-name resolution enforced but cannot prompt without a TTY
> [✗] exit status 125
> make: *** [Makefile:51: linux] Error 1
> ```

I made some changes to detect docker, and if "docker" is a "podman" alias. So I can use "docker" or "podman" command now. I want to fix the ldflags problem before.

internal/command/docker.go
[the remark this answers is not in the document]
metal3d: GOFLAGS seems to be fixed in https://github.com/fyne-io/fyne-cross/pull/42/commits/aa8002daec9555333f2747f24a63c93d1c092489

I didn't tried with "docker". Note that now, the "podman" command is used even if "docker" command is an alias

metal3d:
> > Thanks again for working on this @metal3d. It is a very nice feature to have and would be nice to see it landed soon. Do you think you could get this rebased on develop? The tests don't want to start and I think the out-of-date branch might partly be the reason.
> > I finally managed to test this out on my new Fedora 34 installation. I had to install `podman-docker` to even get it started. Would it be possible to get it to only use `podman`? After that, I ran into this error:
> > ```
> > ~/go/bin/fyne-cross linux -arch amd64,arm64 -app-id com.github.jacalz.wormhole-gui -icon internal/assets/icon/icon-512.png
> > [i] Target: linux/amd64
> > [i] Cleaning target directories...
> > [✓] "bin" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/bin/linux-amd64
> > [✓] "dist" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/dist/linux-amd64
> > [✓] "temp" dir cleaned: /home/jacob/Git/wormhole-gui/fyne-cross/tmp/linux-amd64
> > [i] Checking for go.mod: /home/jacob/Git/wormhole-gui/go.mod
> > [✓] go.mod found
> > [i] Building binary...
> > USE PODMAN
> > Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
> > Error: error getting default registries to try: short-name resolution enforced but cannot prompt without a TTY
> > [✗] exit status 125
> > make: *** [Makefile:51: linux] Error 1
> > ```
> 
> I made some changes to detect docker, and if "docker" is a "podman" alias. So I can use "docker" or "podman" command now. I want to fix the ldflags problem before.

OK, for the registry, now, Fedora proposes several registries. The image location should be "docker.io/blablabla"  for "moby-engine" (the docker distribution by RH/Fedora" and podman. I should not have this problem because I built the image myself (to fix the entrypoint for podman)

So, the image "url" should be changed to it's complete form.

Jacalz (changes requested):
Sorry for the late review. I've had a lot to do with work related stuff.

I still could not use only `podman`, had to install `podman-docker` to get to this point. If it is possible to egt it to only use `podman` directly, that would be great, but not a necessity, I guess. This is looking better, but I am still getting the following output when trying to use it:
```
/home/jacob/go/bin/fyne-cross windows -arch amd64 -app-id io.github.jacalz.seaborne
[i] Target: windows/amd64
[✓] "bin" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/bin/windows-amd64
[✓] "dist" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/dist/windows-amd64
[✓] "temp" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/tmp/windows-amd64
[i] Checking for go.mod: /home/jacob/Git/seaborne/go.mod
[✓] go.mod found
Error: short-name resolution enforced but cannot prompt without a TTY
[✗] could not package the Fyne app: exit status 125
```

metal3d:
> Sorry for the late review. I've had a lot to do with work related stuff.
> 
> I still could not use only `podman`, had to install `podman-docker` to get to this point. If it is possible to egt it to only use `podman` directly, that would be great, but not a necessity, I guess. This is looking better, but I am still getting the following output when trying to use it:
> 
> ```
> /home/jacob/go/bin/fyne-cross windows -arch amd64 -app-id io.github.jacalz.seaborne
> [i] Target: windows/amd64
> [i] Cleaning target directories...
> [✓] "bin" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/bin/windows-amd64
> [✓] "dist" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/dist/windows-amd64
> [✓] "temp" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/tmp/windows-amd64
> [i] Checking for go.mod: /home/jacob/Git/seaborne/go.mod
> [✓] go.mod found
> Error: short-name resolution enforced but cannot prompt without a TTY
> [✗] could not package the Fyne app: exit status 125
> ```

You should not require to install "podman-docker", I haven't it and it works now with "podman" command if needed.

For the "short" link name, this is a problem. As explained, podman (and the "docker" version of Fedora named "moby-engine") proposes to use others registries, so we need TTY to answer. If the image is already present (for example because I compiled the image myself to test my changes) so it works.

The problem is that it seems you don't have the image on your Fedora (and it's absolutely normal) - The "solution" is: we change the image path to "docker.io/fyne/fyne-cross" ("docker.io" added to the image name, so podman or moby-engine understand where is the repository) - but this has got a bad behaviour: we can't use a locally built image...

I'm trying to find a way to make it possible to have a TTY to answer the question asked by podman/moby... this problem doesn't happen if you use "docker" (because they configured the docker client to use "docker.io" by default).

Or, maybe you have a better idea ?

metal3d:
@Jacalz the "pullImage" function can be changed like in https://github.com/fyne-io/fyne-cross/pull/42/commits/e9a493dde52744784d2117b4387f7ba9e564bf3b so it now avoids to ask which repository to use. Could you please check if it's OK on your Fedora ?

metal3d:
@Jacalz I tried and see this:

```
../../fyne-cross/fyne-cross windows
[i] Target: windows/amd64
[✓] "bin" dir cleaned: /home/metal3d/Projects/fyne-experiments/test-cross/fyne-cross/bin/windows-amd64
[✓] "dist" dir cleaned: /home/metal3d/Projects/fyne-experiments/test-cross/fyne-cross/dist/windows-amd64
[✓] "temp" dir cleaned: /home/metal3d/Projects/fyne-experiments/test-cross/fyne-cross/tmp/windows-amd64
[i] Checking for go.mod: /home/metal3d/Projects/fyne-experiments/test-cross/go.mod
[✓] go.mod found
Resolving "fyneio/fyne-cross" using unqualified-search registries (/etc/containers/registries.conf)
Trying to pull registry.fedoraproject.org/fyneio/fyne-cross:1.1-windows...
Trying to pull registry.access.redhat.com/fyneio/fyne-cross:1.1-windows...
Trying to pull docker.io/fyneio/fyne-cross:1.1-windows...
```
and the image is well pulled, but the "pullImage" seems to not being used at this stage. 

But, for me, it never asks for TTY... Are you up to date with podman ?

Jacalz (changes requested):
Thanks for your work on this. the latest changes seems to make it mostly work without installing `podman-docker`, nice work. Looks like a few changes are part of this PR that are unrelated to this. Did the merge go wrong?

I'm still running into the issue with TTY on my fully updated Fedora 34 installation. Am I missing something with how I should use `podman`? I literally just used what I had installed, no tweaking or other changes done on my part.

I also tried doing `fyne-cross windows -pull` and got an error that the docker executable wasn't available. I tried installing `podman-docker` and got this instead:
```
[✓] go.mod found
chown: changing ownership of '/go': Operation not permitted
[✗] could not package the Fyne app: exit status 1
```
Got a similar error when I tried to run a regular build using fyne-cross as well after having installed podman-docker.

metal3d:
Yes. You need to build images because I made some changes on entrypoint.
With podman, there is no need to change owner because I'm using the
"keepid" option. So, for podman, the new entrypoint skips this stage.

For tty, I will try to understand why there is a different behavior on my
laptop. I didn't tweak anything. I removed the images and it pull images
whitout trying to ask me the registry to use. So maybe there's a cache on
previous answers...

I will try to reset everything on my podman installation to reproduce the
problem.

Jacalz:
I did use it thorough a Makefile if I remember correctly. It shouldn't cause any issues, I think, but I'll try without just to be sure.

metal3d:
> Thanks for your work on this. the latest changes seems to make it mostly work without installing `podman-docker`, nice work. Looks like a few changes are part of this PR that are unrelated to this. Did the merge go wrong?

There were something weird as soon as I pull "develop" branch then rebased mine. I don't know why but some things were "recommitted" without any reason.

For example, there is a problem with CHANGELOG.md file... whatever I do, there is a difference between mine and the one from develop branch...

metal3d:
I made fixes on the rebase, it seems that it's now a bit better.
I will try to see how to avoid the need of TTY - it's not that easy to find where you defined the image names (and so URLs) 😉

metal3d:
OK @Jacalz I see the problem.

My laptops have this file: `/etc/containers/registries.conf`, wich is provided by the package named `containers-common`.

(You can check if your configuration is changed from the RPM version `sudo rpm -Vc containers-common` and try `sudo rpmconf -a` to say if you want to take the maintainer version of configuration.)

Inside, there is a list of `unqualified-search-registries` that I didn't touch (I didn't touch this file anyway). Podman uses this list to search images if the "domain" is not set.

When TTY is not accessible, podman searches inside these registries - but not on your laptop. It's weird because I made tests on my 2 laptops and both don't ask anything. But, if you've got the problem, so others will have the same problem.

There are 2 possibilities:

- the first one is to add this line in `docker.go` in the `Cmd` function (line 124): `pullImage(Context{DockerImage: "docker.io/" + image, Pull: true})` - this will pull the image with the right domain - the problem is that it will overwrite the local images, so it breaks local changes on images
- the second is to add an option for "pullImage" function to say "pull if not present" (exactly like Kubernetes does) - this will pull the images if it's not present on the user's computer

The second option is better IMHO, but to needs a bit more modifications. I can do it. Do you want me to make this changes ?

metal3d:
@Jacalz I found a way to avoid the problem in https://github.com/fyne-io/fyne-cross/pull/42/commits/a88ea7999a1b2589f23699815f71737561e2377a

- the makefile tags images with the full name (domain as prefix)
- I add the domain to the image in command calls

So:
- if the user hasn't the image (e.g. it hasn't built them) so the pull is made with the full image name
- if the user has built the images with podman (see the Makefile) so the image is present

This fixes problems for me. It works with Docker and Podman. Note that OCI recommends using the full image name with domain since there are now a few public registries, using the full name avoids image spoofing.

Jacalz (commented):
Thanks for the great work with the changes. I still need to test on my workstation, but I just tested on my laptop (of which I had never used docker or podman). I downloaded this branch and just ran the following command and got the below output.This is getting very close now :)
```
[jacob@fedora seaborne]$ make windows
/home/jacob/go/bin/fyne-cross windows -arch amd64 -app-id io.github.jacalz.seaborne
[i] Target: windows/amd64
[✓] "temp" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/tmp/windows-amd64
[✓] "bin" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/bin/windows-amd64
[✓] "dist" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/dist/windows-amd64
[i] Checking for go.mod: /home/jacob/Git/seaborne/go.mod
[✓] go.mod found
Trying to pull docker.io/fyneio/fyne-cross:1.1-windows...
Getting image source signatures
Copying blob 65d943ee54c1 done  
Copying blob 8962bc0fad55 done  
Copying blob e8d62473a22d done  
Copying blob d960726af2be done  
Copying blob 186c77a2a533 done  
Copying blob f2253e6fbefa done  
Copying blob db807893dccf done  
Copying blob 292f24f0ff89 done  
Copying blob 974bf8135e20 done  
Copying blob 6f6e9bee4d1e done  
Copying blob bf3ad36f9777 done  
Copying blob 2411a53787f3 done  
Copying config dcf827d6fe done  
Writing manifest to image destination
Storing signatures
chown: changing ownership of '/go': Operation not permitted
[✗] could not package the Fyne app: exit status 1
make: *** [Makefile:28: windows] Error 1
```

Jacalz (commented):
Just tested on my workstation and got the same issue with changing ownership. Seems to be working very fine up to that point though :)

I just have to say how much I like this change. Fedora seems to have podman installed by default and getting fyne-cross up and running was as simple as a `go install` without installing anything else (because I had Go installed already) :100:

metal3d:
@Jacalz are we OK that you built images locally ? One more time, my changes needs new images with my modified entrypoint because I use the "keepid" argument, so your "chmod" or "chown" in the entry point will fail:  https://github.com/fyne-io/fyne-cross/blob/a88ea7999a1b2589f23699815f71737561e2377a/docker/base/docker-entrypoint.sh#L4

metal3d:
> Just tested on my workstation and got the same issue with changing ownership. Seems to be working very fine up to that point though :)
> 
> I just have to say how much I like this change. Fedora seems to have podman installed by default and getting fyne-cross up and running was as simple as a `go install` without installing anything else (because I had Go installed already)

I'm very happy to help :)

Yes, podman is the next major container runner, that's why I propose these changes. Fedora (and other distributions that are more regarding licenses) insist on using Podman instead. There are some habits to changes but it's mainly the same method to launch and use containers. Podman is a bit more simple to bind the user id to the container (that's why I changed the entry point). There are a few incompatibilities but it's commonly easy to fix them ;)

Jacalz:
Ah, sorry. I must have misunderstood what you meant by that. Will it work without building locally once the new images are uploaded to docker hub? Will try again as soon as I can.

metal3d:
> > @Jacalz are we OK that you built images locally ? One more time, my changes needs new images with my modified entrypoint because I use the "keepid" argument, so your "chmod" or "chown" in the entry point will fail:  https://github.com/fyne-io/fyne-cross/blob/a88ea7999a1b2589f23699815f71737561e2377a/docker/base/docker-entrypoint.sh#L4
> 
> Ah, sorry. I must have misunderstood what you meant by that. Will it work without building locally once the new images are uploaded to docker hub?
> 
> I will re-test this as son as I have the time. Again, many thanks for this work

Yes it will work because the entrypoint is changed and injected in image. So when you'll push them to the Docker registry, the entrypoint will be there. 

The changes I made are only to avoid `chown` commands when the `use_podman` environment is set. And I add this argument when I detect podman usage: https://github.com/metal3d/fyne-cross/blob/propose-podman-fixes/internal/command/docker.go#L105

Jacalz (changes requested):
Sorry for my stupidity with not understanding that I had to rebuild the images. I finally managed to try and build them, but I got this error:
```
/bin/sh: -c: line 1: syntax error near unexpected token `('
/bin/sh: -c: line 1: `podman build -f /home/jacob/Downloads/fyne-cross-propose-podman-fixes(1)/fyne-cross-propose-podman-fixes/docker/base/Dockerfile -t fyneio/fyne-cross:"1.1"-base .'
make: *** [Makefile:8: base] Error 2
```

metal3d:
> Sorry for my stupidity with not understanding that I had to rebuild the images. I finally managed to try and build them, but I got this error:
> 
> ```
> /bin/sh: -c: line 1: syntax error near unexpected token `('
> /bin/sh: -c: line 1: `podman build -f /home/jacob/Downloads/fyne-cross-propose-podman-fixes(1)/fyne-cross-propose-podman-fixes/docker/base/Dockerfile -t fyneio/fyne-cross:"1.1"-base .'
> make: *** [Makefile:8: base] Error 2
> ```

I imagine that the parenthesis in the directory name are the problem. I should happen with docker also. I can fix the Makefile to avoid this

Jacalz (commented):
Thanks. This is working very well now it seems. I can't run the binaries at this point as I'm on vacation and don't have access to anything other than my Linux laptop currently. Building the images for Windows, Linux and FreeBSD worked great.

When I then ran the Windows build, I had the option to choose where I wanted to grab the Go 1.16.5 base from. I choose docker.io over other redhat, fedora and some other option being avaliable. I don't know if it was the right option to use, but the windows binary then seems to have built without anyy issues at all.

However, I seem to be getting this when building for linux:
```
jacob@fedora seaborne]$ /home/jacob/go/bin/fyne-cross linux -arch amd64 -app-id io.github.jacalz.seaborne
[i] Target: linux/amd64
[✓] "temp" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/tmp/linux-amd64
[✓] "bin" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/bin/linux-amd64
[✓] "dist" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/dist/linux-amd64
[i] Checking for go.mod: /home/jacob/Git/seaborne/go.mod
[✓] go.mod found
[i] Building binary...
error: failed switching to "1000": operation not permitted
[✗] exit status 1
```

Also, FYI, it seems like the CI test are failing.

metal3d:
> Thanks. This is working very well now it seems. I can't run the binaries at this point as I'm on vacation and don't have access to anything other than my Linux laptop currently. Building the images for Windows, Linux and FreeBSD worked great.
> 
> When I then ran the Windows build, I had the option to choose where I wanted to grab the Go 1.16.5 base from. I choose docker.io over other redhat, fedora and some other option being avaliable. I don't know if it was the right option to use, but the windows binary then seems to have built without anyy issues at all.
> 
> However, I seem to be getting this when building for linux:
> 
> ```
> jacob@fedora seaborne]$ /home/jacob/go/bin/fyne-cross linux -arch amd64 -app-id io.github.jacalz.seaborne
> [i] Target: linux/amd64
> [i] Cleaning target directories...
> [✓] "temp" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/tmp/linux-amd64
> [✓] "bin" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/bin/linux-amd64
> [✓] "dist" dir cleaned: /home/jacob/Git/seaborne/fyne-cross/dist/linux-amd64
> [i] Checking for go.mod: /home/jacob/Git/seaborne/go.mod
> [✓] go.mod found
> [i] Building binary...
> error: failed switching to "1000": operation not permitted
> [✗] exit status 1
> ```
> 
> Also, FYI, it seems like the CI test are failing.

I don't understand when you had to choose "docker.io". That never happens here. The problem was when you built for Windows ?

I will retry.

Jacalz:
I only got that once, the first time I built for Windows. I don't know if it really was any problem, the Windows build generally seemed to worked great.

metal3d:
> I only got that once, the first time I built for Windows. I don't know if it really was any problem, the Windows build generally seemed to worked great.

OK so the problem to "switch to user 1000" is because the image is pulled (instead of using the local one), so this is not the local image. The windows image is either not well tagged or named somewhere, or the image is still pulled without the "docker.io" prefix. So... I will retake a look.

metal3d:
> > I only got that once, the first time I built for Windows. I don't know if it really was any problem, the Windows build generally seemed to worked great.
> 
> OK so the problem to "switch to user 1000" is because the image is pulled (instead of using the local one), so this is not the local image. The windows image is either not well tagged or named somewhere, or the image is still pulled without the "docker.io" prefix. So... I will retake a look.

I don't understand...
```

FORCE_PODMAN=true ../fyne-cross/fyne-cross windows -debug
[i] Target: windows/amd64
command.Context{Volume:volume.Volume{binDirHost:"/home/pafer/Projects/dlimage/fyne-cross/bin", cacheDirHost:"/home/pafer/.cache/fyne-cross", distDirHost:"/home/pafer/Projects/dlimage/fyne-cross/dist", tmpDirHost:"/home/pafer/Projects/dlimage/fyne-cross/tmp", workDirHost:"/home/pafer/Projects/dlimage"}, Architecture:"amd64", Env:[]string{"GOOS=windows", "GOARCH=amd64", "CC=x86_64-w64-mingw32-gcc"}, ID:"windows-amd64", LdFlags:[]string{"-H=windowsgui"}, OS:"windows", Tags:[]string(nil), AppBuild:"1", AppID:"", AppVersion:"1.0", CacheEnabled:true, DockerImage:"fyneio/fyne-cross:1.1-windows", Icon:"/home/pafer/Projects/dlimage/Icon.png", Name:"dlimage.exe", Package:".", Release:false, StripDebug:true, Debug:true, Pull:false, Category:"", Certificate:"", Developer:"", Keystore:"", KeystorePass:"", KeyPass:"", Password:"", Profile:""}
[✓] "bin" dir cleaned: /home/pafer/Projects/dlimage/fyne-cross/bin/windows-amd64
[✓] "dist" dir cleaned: /home/pafer/Projects/dlimage/fyne-cross/dist/windows-amd64
[✓] "temp" dir cleaned: /home/pafer/Projects/dlimage/fyne-cross/tmp/windows-amd64
[i] Checking for go.mod: /home/pafer/Projects/dlimage/go.mod
[✓] go.mod found
/usr/bin/podman run --rm -t -w /app/fyne-cross/tmp/windows-amd64 -v /home/pafer/Projects/dlimage:/app:z --userns keep-id -e use_podman=1 -e CGO_ENABLED=1 -e GOCACHE=/go/go-build -e fyne_uid=1000 docker.io/fyneio/fyne-cross:1.1-windows /usr/local/bin/gowindres -arch amd64 -output dlimage.exe -workdir /app/fyne-cross/tmp/windows-amd64
[i] Building binary...
/usr/bin/podman run --rm -t -w /app -v /home/pafer/Projects/dlimage:/app:z --userns keep-id -e use_podman=1 -v /home/pafer/.cache/fyne-cross:/go:z -e CGO_ENABLED=1 -e GOCACHE=/go/go-build -e GOOS=windows -e GOARCH=amd64 -e CC=x86_64-w64-mingw32-gcc -e GOFLAGS=-ldflags=-H=windowsgui -ldflags=-w -ldflags=-s -e fyne_uid=1000 docker.io/fyneio/fyne-cross:1.1-windows go build -o /app/fyne-cross/bin/windows-amd64/dlimage.exe -v .
[✓] Binary: /home/pafer/Projects/dlimage/fyne-cross/bin/windows-amd64/dlimage.exe
[i] Packaging app...
[✓] Package: /home/pafer/Projects/dlimage/fyne-cross/dist/windows-amd64/dlimage.exe.zip
```

It never pull image. I built them with "make build-image" but maybe you should try `make build RUNNER=podman` if you installed `podman-docker` - so you need to force the RUNNER variable to force the "podman-tag" directive and ensure that images are all tagged with "docker.io" prefix

Jacalz:
Hmm, I thought that I didn't have `podman-docker` installed, but I'll have to double-check. I built the images using `make windows  linux`.

metal3d:
> Hmm, I thought that I didn't have `podman-docker` installed, but I'll have to double-check. I built the images using `make windows linux`.

Yes, OK, so you didn't have the tagged images with prefixes. You need to tag both images with "docker.io/" prefix, or... I change the full Makefile to always add the prefix...

Note: I fixed the tests, sorry for this

Jacalz:
It might be useful to have all the sub commands tagged as well, but I'll try again with them tagged.

Jacalz (commented):
I built the windows and linux images and had them tagged afterwards. I then went on building my application for both those platform and it went perfectly. Nice work :)

It would be great if it could automatically tag the images even when building using using `make windows` etc, but I think this generally is soon to be approved from my part (just need to test macOS and FreeBSD when I get back to my desktop computer).

Jacalz:
As this seems to be working very well from my point of view at the moment, it would be great if you could do a more in-depth review @lucor :)

metal3d:
> I built the windows and linux images and had them tagged afterwards. I then went on building my application for both those platform and it went perfectly. Nice work :)
> 

yes ! thanks :)

> It would be great if it could automatically tag the images even when building using using `make windows` etc, but I think this generally is soon to be approved from my part (just need to test macOS and FreeBSD when I get back to my desktop computer).

I will make some tests on my OSX VM later, Just to be sure. I will try to create a FreeBSD VM also (later).


PS: so happy that the PR is numbered 42 😆

internal/command/docker.go
```diff
@@ -38,6 +62,34 @@ type Options struct {
 	Debug        bool     // Debug if true enable log verbosity
 }
 
+// initIsPodman returns true if docker is an alias to podman (e.g. via podman-docker).
+func initIsPodman() bool {
```
lucor, line 66: What about to add a function that returns the `engine` in use, and optionally some helper isXXX  functions ? 
Example:
`func engine() (string, error)` where string can be `docker` or `podman`
`func isEngineDocker() bool` 
`func isEnginePodman() bool`
 
This will allow to remove the init function and add support for new engines.

internal/command/docker.go
```diff
@@ -38,6 +62,34 @@ type Options struct {
...
+	// read the "docker" executable to check if it's a podman wrapper
+
+	if os.Getenv("FYNECROSS_FORCE_PODMAN") == "true" {
```
lucor, line 69: I see the pro to force an engine. If you agree I'd move to  a fyne-cross flag like `--engine=<type>` where the possibile values are: `auto`, `docker`, `podman` (default to auto-detect)

internal/command/docker.go
```diff
@@ -137,7 +203,33 @@ func goBuild(ctx Context) error {
 
 	// add ldflags to command, if any
 	if len(ldflags) > 0 {
-		args = append(args, "-ldflags", fmt.Sprintf("'%s'", strings.Join(ldflags, " ")))
+		flags := make([]string, len(ldflags))
```
lucor, line 206: It seems passing multiple flags does not work:
```fyne-cross  linux -ldflags="-X=main.Name=Podman -X=main.Title=Test" --debug```
```
/usr/bin/docker run --rm -t -w /app -v /code/lucor/test:/app:z -v /home/lucor/.cache/fyne-cross:/go:z -e CGO_ENABLED=1 -e GOCACHE=/go/go-build -e GOOS=linux -e GOARCH=amd64 -e CC=gcc -e GOFLAGS= -ldflags=-w -ldflags=-s -e fyne_uid=1000 docker.io/fyneio/fyne-cross:1.1-base go build -ldflags=-X=-X=main.Title=Test -w -s -o /app/fyne-cross/bin/linux-amd64/test -v .
flag provided but not defined: -w
usage: go build [-o output] [build flags] [packages]
Run 'go help build' for details.
[✗] exit status 2
```
If you agree, I'd add some tests on this part of logic to ensure we won't break any previous command relying on the fyne-cross 'ldflags` or GOFLAGS env variable.

**Review by lucor — CHANGES_REQUESTED** [review]
Thanks for the hard work @metal3d, I think we are almost there.
Sadly I was not able to test directly the podman engine since I'm on Ubuntu 20.04 :(

File: internal/command/docker.go (hunk 6 above)
Comment (metal3d) [reply internal/command/docker.go:206]: There are many problems that come along with the modifications, I'm sorry about this.
I will try to fix this, I really don't understand what happens when env vars are passed to Docker and Podman, sometimes it takes the entire string as one argument, sometimes not...

File: internal/command/docker.go (hunk 5 above)
Comment (metal3d) [reply internal/command/docker.go:69]: It's what I wanted to do initially - I will make the adaptations.

File: internal/command/docker.go (hunk 4 above)
Comment (metal3d) [reply internal/command/docker.go:66]: That's a good idea, I will make the changes as soon as I can. I'm back from vacations, I left the PC aside for 3 weeks, sorry :)

File: internal/command/docker.go (hunk 6 above)
Comment (metal3d) [reply internal/command/docker.go:206]: OK, the problem is that:
```
-ldflags="-X=A=B -X=C=D"
```

is parsed as "one" ldflags, there is no split made.
I will try another way to treat this.

Comment (metal3d) [reply internal/command/docker.go:203]: Fixed in https://github.com/fyne-io/fyne-cross/pull/42/commits/ad1b9d7f6b17dbed9f8380a4b398763c606a30ae but I don't understand how to add a test for this

File: internal/command/docker.go (hunk 4 above)
Comment (metal3d) [reply internal/command/docker.go]: Done in  https://github.com/fyne-io/fyne-cross/pull/42/commits/8edf19f844b25e932571c4975900d478ed72a555

**Comment by Jacalz** [comment]
@metal3d Was I supposed to do a re-review on this? I was never requested for a re-review but it looks like you have pushed a few commits since the last time. If I was, I'm sorry that it took me three months to reach to this :I

File: internal/command/docker.go [hunk 7]
```diff
@@ -19,13 +22,19 @@ const (
...
 	// gowindresBin is the path of the gowindres binary into the docker image
 	gowindresBin = "/usr/local/bin/gowindres"
+	// registry is the docker registry to use to pull images
+	registry = "docker.io"
 )
 
 // CheckRequirements checks if the docker binary is in PATH
 func CheckRequirements() error {
-	_, err := execabs.LookPath("docker")
+	_, err := exec.LookPath("docker")
```
Jacalz, line 31: Could you please switch back to `execabs` (applies to more places in this PR)? It is being used because `exec` has a security issue on Windows that can't be fixed without breaking the API stability guarantee.

Jacalz (changes requested):
A few notes on switching to `execabs` instead of `exec`.

lucor:
Thanks @metal3d and sorry for the long wait.

metal3d:
I'm sorry to not have answered for a while. Many things happened last months and I missed a lot of discussion.

I'm very happy you accepted this PR :)

Jacalz:
No worries. Thanks for all your hard work!

fiurthorn:
using `go install github.com/fyne-io/fyne-cross@develop` 
on results in

```
$ fyne-cross windows
[i] Target: windows/amd64
[✓] "bin" dir cleaned: /home/fiurthorn/workspace/fyne/fyne-cross/bin/windows-amd64
[✓] "dist" dir cleaned: /home/fiurthorn/workspace/fyne/fyne-cross/dist/windows-amd64
[✓] "temp" dir cleaned: /home/fiurthorn/workspace/fyne/fyne-cross/tmp/windows-amd64
[i] Checking for go.mod: /home/fiurthorn/workspace/fyne/go.mod
[✓] go.mod found
chown: changing ownership of '/go': Operation not permitted
[✗] could not package the Fyne app: exit status 1
```

```
$ podman version
Version:      3.2.1
API Version:  3.2.1
Go Version:   go1.16.7
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=21.10
DISTRIB_CODENAME=impish
DISTRIB_DESCRIPTION="Ubuntu 21.10"
```

Jacalz:
You can't use it just yet with as the updated images aren't on docker hub currently (I think so at least). They need to be tagged differently, so you'll either have to wait for new images or run `make build-images` (takes a long time) locally on your computer.

PR #43, opened 2021-04-26
Fix app-id flag regression

lucor:
This PR update the package and release command to not use the appID value if empty.
This applies only for target OSes that does not require it.


andydotxyz (approved):
Good catch. I wonder if perhaps the commands output should have more text coverage so we could cover this?

lucor:
> Good catch. I wonder if perhaps the commands output should have more text coverage so we could cover this?

Agree. We'd have more text coverage on the output command. 
Unfortunately it requires some refactoring to split the function in order to have a `creation` and `run` functions like the `exec` package does. But I'd move this for a later step, since there is a possibility to get rid completely of the docker binary dep in favour of a Go client library to support multiple container engines.

PR #51, opened 2021-06-07
Update Go to v1.16.5

lucor:
This PR updates Go to v1.16.5

Additionally merges master (v1.1.0) into develop.


andydotxyz:
Looks good, but is it missing a Changelog entry to mention the new version (I just noticed as the Changelog was updated to 1.16.4 by the master merge...

lucor:
It has been updated only into the docker/CHANGELOG.md file.
Wondering if we should move the docker stuff to a dedicated project / repo (i.e. fyne-cross-images) since it is has not impact on the fyne-cross versioning and has a separated lifecycle. Thoughts?

andydotxyz:
> Wondering if we should move the docker stuff to a dedicated project / repo (i.e. fyne-cross-images) since it is has not impact on the fyne-cross versioning and has a separated lifecycle. Thoughts?

That may be worth exploring, if it will keep history cleaner and make it easier to manage.

andydotxyz (approved):

PR #63, opened 2021-08-19
Update docker images: Go v1.17 and Android builds SDK 30 

lucor:
This PR updates the `android` docker image to use the latest command line tools and updates the Android builds to SDK 30

### DIFF
--- a/docker/android/Dockerfile
+++ b/docker/android/Dockerfile
@@ -3,11 +3,11 @@ FROM fyneio/fyne-cross:${FYNE_CROSS_VERSION}-base
 
 ENV JAVA_HOME /usr/local/android_jdk8
 ENV ANDROID_HOME /usr/local/android_sdk
-ENV ANDROID_SDK_TOOLS_VERSION 4333796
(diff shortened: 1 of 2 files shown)


Jacalz:
I'm wondering if we perhaps should wait a little bit more before jumping onto 1.17. Not that I have used it yet and/or found any breaking bugs, but it might be worth waiting for one or two bug fixes before we make the change. It also lets the rest of the ecosystem catch up a bit (if it needs to).

It would also be nice if we could get Windows/arm64 support in but I don't have any machine to run that so I'll not personally shout it out as a priority :)

andydotxyz (dismissed):
Cool :)

andydotxyz (approved):
Yeah, Jacob's point is a good one

lucor:
@Jacalz it is a good point. Yes, the idea was to add support for windows/arm64 as well but let's remove the Go 1.17 update from this PR (done in 0615997) and handle later with a dedicated one :)

Jacalz (approved):
Sounds good 👍

PR #64, opened 2021-08-20
Support specifying target architectures for Android

### LINKED ISSUE #52: Support specifying target architectures for Android
I have a 64-bit only application that I'd like to build for Android. Currently, fyne-cross only supports building for all architectures on Android, so my app fails to build.

Since the desktop builders already have an `-arch` flag, I think it would be natural to support the same for Android.

lucor:
This PR  adds the possibility to specify a target architecture for Android via the `-arch` flag.
It also introduces the universal architecture constant.

Fixes #52


andydotxyz:
I wonder if it needs to be more specific about what universal is - the docs simply say multiple architectures.
Or maybe not, I guess this is platform specific.
Does it need to be applied to iOS and macOS as well?

lucor:
> I wonder if it needs to be more specific about what universal is - the docs simply say multiple architectures.
> Or maybe not, I guess this is platform specific.
Good point. I'll update to use multiple architectures.

> Does it need to be applied to iOS and macOS as well?
That would be great. Wondering if we should implement this feature into the fyne cli tool.

Jacalz:
Sorry @lucor. I don't have any Android phone to actually test this on.

andydotxyz:
> That would be great. Wondering if we should implement this feature into the fyne cli tool.

We can look at it. Let's chat to @changkun who is working on many elements of Fyne iOS just now :)

andydotxyz (approved):
Great, nice addition

internal/command/context.go
```diff
@@ -22,6 +22,9 @@ const (
 	ArchArm64 Architecture = "arm64"
 	// ArchArm represents the arm architecture
 	ArchArm Architecture = "arm"
+	// ArchMultiple represents the universal architecture used by some OS to
+	// identify a binary that supports multiple architectures (fat binary)
+	ArchMultiple Architecture = "multiple"
```
changkun, line 27: How about calling this "universal"?

internal/command/android.go
```diff
@@ -41,8 +46,10 @@ func (cmd *Android) Parse(args []string) error {
 
 	flags := &androidFlags{
 		CommonFlags: commonFlags,
+		TargetArch:  &targetArchFlag{string(ArchMultiple)},
 	}
 
+	flagSet.Var(flags.TargetArch, "arch", fmt.Sprintf(`List of target architecture to build separated by comma. Supported arch: %s.`, androidArchSupported))
```
changkun, line 52: Is it better to have one more `-arch` option? How about, for example `-os android/amd64` where `-os` consumes `OS/ARCH`?

It can also parse multiple targets, such as: `-os android/386,amd64,arm,arm64`, or just `-os android/universal`.

changkun (approved):
iOS seems much simpler than this, as nowadays it only has `ios/arm64` as a combination :)

Although I recently proposed to add one more OS target `iossimulator` to address fat binary issue for darwin/arm64, it would be nicer to not extend a `-arch` but unify them in `-os` where takes arguments that can optionally specify target architecture `-os OS[/ARCH1,ARCH2,...]` (IMO).

changkun:
(sorry I didn't meant "approve", but just want to write some comments..)

internal/command/context.go
lucor: it was the initial proposal, but we changed to multiple architecture to align with the android docs as pointed out by Andrew. 
Probably if we want to uniform to the "universal" term across android and iOS / macOS this is a good time :)

andydotxyz: I'm not hugely fussed either way - it was mostly that the docs need to be clear, the concept of universal varies across platforms and versions so we need to be careful how it is described.
Universal, in my mind at least, means "all architectures" which may or not be what is built under the new feature.

internal/command/android.go
lucor: That makes sense. The choice to have two options `arch` and `os` was added to follow the Go `GOARCH` and `GOOS`. We'd could change them but could introduce a potential breaking change.
Also please note that the fyne cli and the fyne-cross cli are slightly different and would great if in the future we'd align them.

lucor:
Going to merge, created https://github.com/fyne-io/fyne/issues/2515 to keep track of macOS / iOS discussion.

PR #80, opened 2021-10-07
fixes darwin image creation (SDK extraction)

alexadhy:
### Description:
Fixes darwin image creation (SDK is using xz instead of bz2 compression)

### Checklist:

- [X] Lint and formatter run with no errors.
- [x] Tests all pass.


andydotxyz:
Maybe it should somehow be `.tar.*` so it supports both formats? (Assuming it changed recently and users may be downloading both?).
It would be a more complicated change but feels more complete.

Issue #98, opened 2022-04-01
Add Kubernetes cluster support


Bluebugs:
### Is your feature request related to a problem? Please describe:

It is possible to setup a cluster of machine that will be able to run multiple containers at the same time or sequentially using Kubernetes based infrastructure. Technically, it should be possible to reuse most of the existing logic in fyne-cross to drive a cluster to do the build work.

### Is it possible to construct a solution with the existing API?

Specifying the cluster is done implicitly with the kubernetes config file. What could be necessary and trigger the switch to use kubernetes could be the specification of a namespace to use with the '-n' parameter usually used by kubectl command.

### Describe the solution you'd like to see:
This is still a reflection and not something where I am sure on how things need to be done.

A little bit of background, kubernetes has an API which orchestrate things in the cluster. This has to use it, but this API is low bandwidth. We shouldn't for example transfer file through it in large quantity. It is ok for debug purpose, but not for production. The container in a kubernetes cluster are named pods. They have ressource attached to them, like filesystem and config. The pods are what fyne-cross should control.

Each machine that form a cluster is named a node and with certain configuration, they can scale on demand too. That's something fyne-cross shouldn't try to control.

You can experiment locally with tool like minikube.

The main difference between podman/docker backend and kubernetes backend is that the container is not going to run on the same machine. Which means all the necessary data must be send over to the container and not rely on the availability of the local file system.

My current idea for that purpose is that all the mounting point are packaged with tar zstd and uploaded to a S3 bucket by fyne-cross. This is usually available has a public front service and can be mimicked with open source solution like min.io. Then the pod during initialization is pointed to that S3 bucket and locally unpack the content. This allow for uploading things once and reusing it potentially for multiple build in a row. Once a build is done, the resulting artifact would also be uploaded to S3. fyne-cross could then download them back on the local machine.

All the configuration for this cloud services are usually stored in environment variable or yaml file. We might start by just supporting environment variable. I will try to build a prototype of this shortly to see how usable it is.

metal3d:
Actually, the best practice is to use CI/CD tools to make this. And Kubernetes can use "Jobs" to make what you expect too but that's not a good practice. I already have made this with Gitlab runners and Drone.io that builds my apps on Kubernetes.
[...]

PR #87, opened 2022-01-10
Add support for FyneApp.toml

lucor:
### Description:

This PR adds support for the  the FyneApp.toml file

Fixes #78

### Checklist:

- [x] Tests included.
- [x] Lint and formatter run with no errors.
- [x] Tests all pass.

Issue #85, opened 2021-12-30
fyne-cross fails when trying to create TAR file.


johnnewcombe:
### Describe the bug:

When running the command...

`fyne-cross linux -pull -app-id=com.glasstty.telstar-client -arch=amd64,arm64`

I receive the following error:

```
Failed to create archive with tar: exit status 1
[✗] could not package the Fyne app: could not package the Fyne app: exit status 1
```
The binary itself is created successfully.

All builds and packaging work as expected for Windows (386/amd64) and Darwin (amd64 and arm64).

The docker images I am using are:

```
fyneio/fyne-cross        1.1-windows       877befb40c44   3 weeks ago    1.62GB
fyneio/fyne-cross        1.1-linux-arm64   05adcb22a5ba   3 weeks ago    1.48GB
fyneio/fyne-cross        1.1-base          7e37d8c376af   3 weeks ago    1.07GB 
```
### Device and debug info (please complete the following information):

<summary>Device info</summary>

 - **OS:** MacOS
 - **Version:**  11.6
 - **Go version:** go version go1.17.1 darwin/amd64
 - **fyne-cross version:** v1.1.3
 - **Fyne version:** fyne cli version: v2.1.2

<details><summary>Debug info</summary><br><pre>

$ fyne-cross linux -pull -debug -no-cache -app-id=com.glasstty.telstar-client -arch=amd64,arm64

....
github.com/yuin/goldmark/renderer
github.com/yuin/goldmark/parser
github.com/yuin/goldmark/renderer/html
github.com/yuin/goldmark
fyne.io/fyne/v2/widget
fyne.io/fyne/v2/container
fyne.io/fyne/v2/dialog
bitbucket.org/johnnewcombe/telstar-client/display
fyne.io/fyne/v2/internal/painter/gl
fyne.io/fyne/v2/internal/driver/common
fyne.io/fyne/v2/internal/driver/glfw
fyne.io/fyne/v2/app
bitbucket.org/johnnewcombe/telstar-client
[✓] Binary: /Users/john/Development/Repositories/Telstar/telstar-client/fyne-cross/bin/linux-amd64/telstar-client
[i] Packaging app...
/usr/local/bin/docker run --rm -t -w /app -v /Users/john/Development/Repositories/Telstar/telstar-client:/app -e CGO_ENABLED=1 -e GOCACHE=/go/go-build -e fyne_uid=501 fyneio/fyne-cross:1.1-base /usr/local/bin/fyne version
[...]

Issue #46, opened 2021-05-17
parameters and their documentation are inconsistent


andydotxyz:
I just noticed that if you omit `-app-id` it says that `appID` is required. This led me to notice a few inconsistencies that should be tidied up to avoid confusion.

ManuelReschke:
I wanted to build biniaries for android but i got this message:

fyne-cross android -arch=arm,arm64
[✗] appID is mandatory for android

What is the APP ID ? I thougt it was that????
-> myApp := app.NewWithID(AppID)

But it wasnt. Can someone explain what it is?

andydotxyz:
The packaging needs an appID for android and iOS. The one in code is for Fyne storage etc.
We have reconciled them to both draw from `FyneApp.toml` but if specified manually they are still separate and both are required (https://developer.fyne.io/started/metadata).

hhirsch:
@andydotxyz thanks to you I was able to fix my issue.
Just in case people come here by search engine and it is not clear enough:
When building for windows with `fyne-cross windows -arch=amd64,386` I also had to create FyneApp.toml

andydotxyz:
I think this also relates to the tools migration @sdassow

PR #101, opened 2022-04-06
Refactor and reuse code

Bluebugs:
### Description:
This is "just" a refactor. As I was working on #98, I realize it would be simpler if there was a bit more potential for reusing code around. I am not sure that all of those change are really idiomatic go and feel like doing a PR that "just" move code around and remove duplicated step will make the next PR easier to move forward.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:
There should be no functional change. I have only the ability to test with podman at the moment and will be glad if anyone tried with docker.

### DIFF
--- a/internal/command/android.go
+++ b/internal/command/android.go
@@ -24,17 +24,12 @@ var (
 
 // Android build and package the fyne app for the android OS
 type Android struct {
-	Context []Context
+	CrossBuilderCommand
+	CrossBuilder
 }
 
-// Name returns the one word command name
-func (cmd *Android) Name() string {
-	return "android"
-}
-
-// Description returns the command description
-func (cmd *Android) Description() string {
-	return "Build and package a fyne application for the android OS"
+func NewAndroidCommand() *Android {
+	return &Android{CrossBuilder: CrossBuilder{name: "android", description: "Build and package a fyne application for the android OS"}}
 }
 
 // Parse parses the arguments and set the usage for the command
@@ -57,101 +52,68 @@ func (cmd *Android) Parse(args []string) error {
 	flagSet.Usage = cmd.Usage
 	flagSet.Parse(args)
 
-	ctx, err := makeAndroidContext(flags, flagSet.Args())
+	err = cmd.makeAndroidContainerImages(flags, flagSet.Args())
+	return err
+}
+
+// Run runs the command using helper code
+func (cmd *Android) Run() error {
+	return cmd.RunInternal(cmd)
+}
+
+// Run runs the command
+func (cmd *Android) RunEach(image ContainerImage) error {
+	//
+	// package
+	//
+	log.Info("[i] Packaging app...")
+
+	packageName := fmt.Sprintf("%s.apk", cmd.defaultContext.Name)
+
+	err := prepareIcon(cmd.defaultContext, image)
 	if err != nil {
 		return err
 	}
 
-	cmd.Context = ctx
-	return nil
-}
-
-// Run runs the command
-func (cmd *Android) Run() error {
-
-	for _, ctx := range cmd.Context {
-
-		err := bumpFyneAppBuild(ctx)
-		if err != nil {
-			log.Infof("[i] FyneApp.toml: unable to bump the build number. Error: %s", err)
-		}
-
-		log.Infof("[i] Target: %s/%s", ctx.OS, ctx.Architecture)
-		log.Debugf("%#v", ctx)
-
-		//
-		// pull image, if requested
-		//
-		err = pullImage(ctx)
-		if err != nil {
-			return err
-		}
-
-		//
-		// prepare build
-		//
-		err = cleanTargetDirs(ctx)
-		if err != nil {
-			return err
-		}
-
-		err = goModInit(ctx)
-		if err != nil {
-			return err
-		}
-
-		//
-		// package
-		//
-		log.Info("[i] Packaging app...")
-
-		packageName := fmt.Sprintf("%s.apk", ctx.Name)
-
-		err = prepareIcon(ctx)
-		if err != nil {
-			return err
-		}
-
-		if ctx.Release {
-			err = fyneRelease(ctx)
-		} else {
-			err = fynePackage(ctx)
-		}
-		if err != nil {
-			return fmt.Errorf("could not package the Fyne app: %v", err)
-		}
-
-		// move the dist package into the "dist" folder
-		// The fyne tool sanitizes the package name to be acceptable as a
-		// android package name. For details, see:
-		// https://github.com/fyne-io/fyne/blob/v1.4.0/cmd/fyne/internal/mobile/build_androidapp.go#L297
-		// To avoid to duplicate the fyne tool sanitize logic here, the location of
-		// the dist package to move will be detected using a matching pattern
-		apkFilePattern := volume.JoinPathHost(ctx.WorkDirHost(), ctx.Package, "*.apk")
-		apks, err := filepath.Glob(apkFilePattern)
-		if err != nil {
-			return fmt.Errorf("could not find any apk file matching %q: %v", apkFilePattern, err)
-		}
-		if apks == nil {
-			return fmt.Errorf("could not find any apk file matching %q", apkFilePattern)
-		}
-		if len(apks) > 1 {
-			return fmt.Errorf("multiple apk files matching %q: %v. Please remove and build again", apkFilePattern, apks)
-		}
-		srcFile := apks[0]
-		distFile := volume.JoinPathHost(ctx.DistDirHost(), ctx.ID, packageName)
-		err = os.MkdirAll(filepath.Dir(distFile), 0755)
-		if err != nil {
-			return fmt.Errorf("could not create the dist package dir: %v", err)
-		}
-
-		err = os.Rename(srcFile, distFile)
-		if err != nil {
-			return err
-		}
-
-		log.Infof("[✓] Package: %s", distFile)
+	if cmd.defaultContext.Release {
+		err = fyneRelease(cmd.defaultContext, image)
+	} else {
+		err = fynePackage(cmd.defaultContext, image)
 	}
+	if err != nil {
+		return fmt.Errorf("could not package the Fyne app: %v", err)
+	}
+
+	// move the dist package into the "dist" folder
+	// The fyne tool sanitizes the package name to be acceptable as a
+	// android package name. For details, see:
+	// https://github.com/fyne-io/fyne/blob/v1.4.0/cmd/fyne/internal/mobile/build_androidapp.go#L297
+	// To avoid to duplicate the fyne tool sanitize logic here, the location of
+	// the dist package to move will be detected using a matching pattern
+	apkFilePattern := volume.JoinPathHost(cmd.defaultContext.WorkDirHost(), cmd.defaultContext.Package, "*.apk")
+	apks, err := filepath.Glob(apkFilePattern)
+	if err != nil {
+		return fmt.Errorf("could not find any apk file matching %q: %v", apkFilePattern, err)
+	}
+	if apks == nil {
+		return fmt.Errorf("could not find any apk file matching %q", apkFilePattern)
+	}
+	if len(apks) > 1 {
+		return fmt.Errorf("multiple apk files matching %q: %v. Please remove and build again", apkFilePattern, apks)
+	}
+	srcFile := apks[0]
+	distFile := volume.JoinPathHost(cmd.defaultContext.DistDirHost(), image.GetID(), packageName)
+	err = os.MkdirAll(filepath.Dir(distFile), 0755)
+	if err != nil {
+		return fmt.Errorf("could not create the dist package dir: %v", err)
+	}
+
+	err = os.Rename(srcFile, distFile)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("[✓] Package: %s", distFile)
 	return nil
 }
 
@@ -190,56 +152,48 @@ type androidFlags struct {
 }
 
 // makeAndroidContext returns the command context for an android target
-func makeAndroidContext(flags *androidFlags, args []string) ([]Context, error) {
+func (cmd *Android) makeAndroidContainerImages(flags *androidFlags, args []string) error {
 
 	targetArch, err := targetArchFromFlag(*flags.TargetArch, androidArchSupported)
 	if err != nil {
-		return []Context{}, fmt.Errorf("could not make build context for %s OS: %s", androidOS, err)
+		return fmt.Errorf("could not make build context for %s OS: %s", androidOS, err)
 	}
 
-	ctxs := []Context{}
+	ctx, err := makeDefaultContext(flags.CommonFlags, args)
+	if err != nil {
+		return err
+	}
+
+	// appID is mandatory for android
+	if ctx.AppID == "" {
+		return fmt.Errorf("appID is mandatory for %s", androidOS)
+	}
+
+	cmd.defaultContext = ctx
+	runner := NewContainerRunner(ctx)
+
 	for _, arch := range targetArch {
-		ctx, err := makeDefaultContext(flags.CommonFlags, args)
-		if err != nil {
-			return []Context{}, err
-		}
-
-		// appID is mandatory for android
-		if ctx.AppID == "" {
-			return []Context{}, fmt.Errorf("appID is mandatory for %s", androidOS)
-		}
-
 		// By default, the fyne cli tool builds a fat APK for all supported
 		// instruction sets (arm, 386, amd64, arm64). A subset of instruction sets can
 		// be selected by specifying target type with the architecture name.
 		// E.g.: -os=android/arm
-		ctx.OS = androidOS
-		ctx.Architecture = arch
-		ctx.ID = androidOS
-		if ctx.Architecture != ArchMultiple {
-			ctx.ID = fmt.Sprintf("%s-%s", ctx.OS, ctx.Architecture)
-		}
+		image := runner.NewImageContainer(arch, androidOS, overrideDockerImage(flags.CommonFlags, androidImage))
 
 		if path.IsAbs(flags.Keystore) {
-			return []Context{}, fmt.Errorf("keystore location must be relative to the project root: %s", ctx.Volume.WorkDirHost())
+			return fmt.Errorf("keystore location must be relative to the project root: %s", ctx.Volume.WorkDirHost())
 		}
 
 		_, err = os.Stat(volume.JoinPathHost(ctx.Volume.WorkDirHost(), flags.Keystore))
 		if err != nil {
-			return []Context{}, fmt.Errorf("keystore location must be under the project root: %s", ctx.Volume.WorkDirHost())
+			return fmt.Errorf("keystore location must be under the project root: %s", ctx.Volume.WorkDirHost())
 		}
 
-		ctx.Keystore = volume.JoinPathContainer(ctx.Volume.WorkDirContainer(), flags.Keystore)
-		ctx.KeystorePass = flags.KeystorePass
-		ctx.KeyPass = flags.KeyPass
+		cmd.defaultContext.Keystore = volume.JoinPathContainer(cmd.defaultContext.Volume.WorkDirContainer(), flags.Keystore)
+		cmd.defaultContext.KeystorePass = flags.KeystorePass
+		cmd.defaultContext.KeyPass = flags.KeyPass
 
-		// set context based on command-line flags
-		if flags.DockerImage == "" {
-			ctx.DockerImage = androidImage
-		}
-
-		ctxs = append(ctxs, ctx)
+		cmd.Images = append(cmd.Images, image)
 	}
 
-	return ctxs, nil
+	return nil
 }
--- a/internal/command/android.go
+++ b/internal/command/android.go
@@ -23,21 +23,21 @@ var (
 )
 
 // Android build and package the fyne app for the android OS
-type Android struct {
-	CrossBuilder
+type android struct {
+	crossBuilder
 }
 
-var _ PlatformBuilder = (*Android)(nil)
-var _ Command = (*Android)(nil)
+var _ platformBuilder = (*android)(nil)
+var _ Command = (*android)(nil)
 
-func NewAndroidCommand() *Android {
-	r := &Android{CrossBuilder: CrossBuilder{name: "android", description: "Build and package a fyne application for the android OS"}}
+func NewAndroidCommand() *android {
+	r := &android{crossBuilder: crossBuilder{name: "android", description: "Build and package a fyne application for the android OS"}}
 	r.builder = r
 	return r
 }
 
 // Parse parses the arguments and set the usage for the command
-func (cmd *Android) Parse(args []string) error {
+func (cmd *android) Parse(args []string) error {
 	commonFlags, err := newCommonFlags()
 	if err != nil {
 		return err
@@ -56,12 +56,12 @@ func (cmd *Android) Parse(args []string) error {
 	flagSet.Usage = cmd.Usage
 	flagSet.Parse(args)
 
-	err = cmd.makeAndroidContainerImages(flags, flagSet.Args())
+	err = cmd.setupContainerImages(flags, flagSet.Args())
 	return err
 }
 
 // Run runs the command
-func (cmd *Android) Build(image ContainerImage) (string, string, error) {
+func (cmd *android) Build(image containerImage) (string, string, error) {
 	//
 	// package
 	//
@@ -106,7 +106,7 @@ func (cmd *Android) Build(image ContainerImage) (string, string, error) {
 }
 
 // Usage displays the command usage
-func (cmd *Android) Usage() {
+func (cmd *android) Usage() {
 	data := struct {
 		Name        string
 		Description string
@@ -139,8 +139,8 @@ type androidFlags struct {
 	TargetArch *targetArchFlag
 }
 
-// makeAndroidContext returns the command context for an android target
-func (cmd *Android) makeAndroidContainerImages(flags *androidFlags, args []string) error {
+// setupContainerImages returns the command context for an android target
+func (cmd *android) setupContainerImages(flags *androidFlags, args []string) error {
 
 	targetArch, err := targetArchFromFlag(*flags.TargetArch, androidArchSupported)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


lucor:
I had a cursory look at PR and I would say that the refactor make sense to me but some thoughts come in mind... :-)
Looking at the amount of code changed I was thinking if it could be the right time to move fyne-cross as command of the fyne CLI app, something like `fyne cross`.
Having it as part of the fyne CLI will bring the following pros:
- remove the [direct dependency](https://github.com/fyne-io/fyne-cross/blob/v1.2.0/docker/base/Dockerfile#L3) from the Docker image. This will reduce the number of releases and possible issues due to versions not aligned (see #85)
- all the fyne CLI APIs can be invoked directly, no need to use the `exec.Cmd` (thinking about the build and package steps here)
- uniform the CLI APIs and flags (see #46)
- the support to FyneApp.toml and build metadata would not require code duplication (see https://github.com/fyne-io/fyne-cross/pull/87#issuecomment-1019441410)

Not strictly related but something could be helpful is to move the Docker image code into a dedicated repository (i.e. `fyne-cross-image`). Right now it they are different projects with a proper life-cycle that shares the same repo.

Thoughts?

Bluebugs:
I personally like the idea of moving fyne-cross to become fyne cross and get synchronized release. It will have to be discussed with the rest of the community. Maybe we should create a set of issues toward that goal in fyne-cross and aim for inclusion of fyne-cross in fyne for 2.3.

Jacalz (approved):
I suspect that @lucor will have more (if anything) to say about the code. It all looks sensible to me but I don't quite have the same knowledge over the codebase. All my local builds of rymdport also work just as before, hence approved :)

andydotxyz:
> could be the right time to move fyne-cross as command of the fyne CLI app, something like `fyne cross`.

This is a great thing to work towards, however I don't think I would have it as a separate command (`fyne cross`) but more likely an extension of the whole command where we improve the cross-compile system and allow the fyne-cross path to be run if/when appropriate.

lucor:
Good point, that makes sense to me. We already do a similar thing when building from a macOS with a darwin target

internal/command/command.go
```diff
@@ -22,6 +22,67 @@ type Command interface {
 	Run() error                // Run runs the command
 }
 
+type CrossBuilderCommand interface {
+	RunEach(image ContainerImage) (string, string, error) // Called to build each possible architecture/OS combination
```
lucor, line 26: Not sure about the `image` name. Should not be something like `container` ?

internal/command/container.go
```diff
@@ -0,0 +1,393 @@
...
+
+const (
+	// fyneBin is the path of the fyne binary into the docker image
+	fyneBin = "/usr/local/bin/fyne"
+	// gowindresBin is the path of the gowindres binary into the docker image
+	gowindresBin = "/usr/local/bin/gowindres"
+)
+
+type ContainerRunner interface {
+	NewImageContainer(arch Architecture, OS string, image string) ContainerImage
```
lucor, line 23: Not sure about having a constructor defined as part of an interface

internal/command/container.go
```diff
@@ -0,0 +1,393 @@
...
+
+	Debug(v ...interface{})
+	GetDebug() bool
```
lucor, line 26: What about moving these methods into a dedicated interface, like `Debugger` ?

internal/command/container.go
```diff
@@ -0,0 +1,393 @@
...
+
+	Env  map[string]string // Env is the list of custom env variable to set. Specified as "KEY=VALUE"
+	Tags []string          // Tags defines the tags to use
+
+	vol volume.Volume
+
+	debug bool
+}
+
+type ContainerImage interface {
```
lucor, line 40: As said above not sure about the name here. Maybe something related to a builder ?

internal/command/container.go
```diff
@@ -0,0 +1,393 @@
...
+	Finalize(srcFile string, packageName string) error
+
+	GetArchitecture() Architecture
+	GetOS() string
+	GetID() string
+	GetTarget() string
+	GetEnv(string) (string, bool)
+	SetEnv(string, string)
+	SetMount(string, string)
+	AppendTag(string)
```
lucor, line 53: Do we really need to add these methods as part of an interface ? Should be they removed in favor of a contract with the `AllContainerImage` struct ?

internal/command/container.go
```diff
@@ -0,0 +1,393 @@
...
+
+	GetRunner() ContainerRunner
+}
+
+type ContainerMountPoint struct {
+	LocalHost   string
+	InContainer string
+}
+
+type AllContainerImage struct {
```
lucor, line 63: Not sure about the naming here. The `All` prefix makes me to think about a collection

lucor (commented):
Had a little in depth look at the code and the most of the feedback are related about the naming like: `ContainerImage`, `AllContainerImage`, `AllContainerRunner`. 

Anyway all changes are `internal` and it works locally using docker as before, so feel free to merge if we need it for #108. I'd not block this with a discussion about naming :) Probably they will be lost during the porting to fyne CLI and in any case we can always address them in to a later stage.

lucor:
@Bluebugs it looks like the `android` command is not available:
```
The commands are:

	darwin-image  Build the docker image for darwin
	darwin        Build and package a fyne application for the darwin OS
	linux         Build and package a fyne application for the linux OS
	windows       Build and package a fyne application for the windows OS
	              
	ios           Build and package a fyne application for the iOS OS
	freebsd       Build and package a fyne application for the freebsd OS
	version       Print the fyne-cross version information

Use "fyne-cross <command> -help" for more information about a command.```

**Comment by Bluebugs** [comment]
> @Bluebugs it looks like the `android` command is not available:

Indeed, good catch. I missed that part. Will fix it shortly.

File: internal/command/command.go (hunk 1 above)
Comment (Bluebugs) [reply internal/command/command.go:26]: I don't know. I choose image as at least it was consistently used by docker/podman/kubernetes. Otherwise kubernetes call the assemblage of image+volume on a nodes a `pod`.

File: internal/command/container.go (hunk 4 above)
Comment (Bluebugs) [reply internal/command/container.go:40]: What about `ImageInstance`?

File: internal/command/container.go (hunk 5 above)
Comment (Bluebugs) [reply internal/command/container.go:53]: That sounds doable. I was thinking I would have to change the behavior of some of them, but actually not at all.

File: internal/command/command.go (hunk 1 above)
Comment (andydotxyz) [reply internal/command/command.go:26]: I am more confused about the `Each`... If this is iterating through architectures, where is that state passed?
I would avoid `Each` if we can be more specific, it works for a language primitive but here seems too generic.
How does this interface relate to the concrete `CrossBuilder`?

File: internal/command/container.go (hunk 2 above)
Comment (andydotxyz) [reply internal/command/container.go:23]: Agreed, that isn't the Go pattern, It is not clear what `ContainerRunner` is describing.

File: internal/command/container.go [hunk 7]
```diff
@@ -0,0 +1,393 @@
...
+)
+
+type ContainerRunner interface {
+	NewImageContainer(arch Architecture, OS string, image string) ContainerImage
+
+	Debug(v ...interface{})
+	GetDebug() bool
+}
+
+type AllContainerRunner struct {
```
andydotxyz, line 29: Is this the top level of Fyne-Cross running? (from the content it seems like it might be)?
If so `CrossCompiler` or something high level but specific would be helpful

internal/command/container.go
andydotxyz: I think there is a peculiar relationship between `ContainerRunner`, `AllContainerRunner` and `ContainerImage` here, solving the overall model might lead to a good name here.
What is the intent of these interfaces/structs?

There is also a circular reference between `ContainerRunner` and `ContainerImage`. I think that this discussion is not so much a question of naming as the grouping itself.

internal/command/container.go
andydotxyz: Remember that in Go it is idiomatic to remove `Get` prefix for simple getter calls.

andydotxyz: Are the Os/Arch part of the API at all? a `ContainerImage` is a generic construct that is created by the `ContainerRunner` calling some constructor passing in state, so this feels a little circular.

internal/command/container.go
```diff
@@ -0,0 +1,393 @@
...
+	return nil
+}
+
+func (a *AllContainerRunner) Debug(v ...interface{}) {
+	if a.debug {
+		log.Debug(v...)
+	}
+}
+
+func (a *AllContainerRunner) GetDebug() bool {
```
andydotxyz, line 88: I see why `Get` is needed here because you have mixed the noun and verb of `Debug`, is this perhaps `DebugEnabled` or even `Debugging`?

andydotxyz:
I have added some thoughts around naming. However I was not able to add suggestions in most places as it is not clear of the intent from model presented. Can you please summarise what each of the new interfaces/structs are supposed to describe in one comment so we can understand the bigger picture @Bluebugs? It is a bit confusing that runners contain images bit images have run methods.

internal/command/command.go
Bluebugs: It is iterating on an image which is hold the information regarding engine/os/architecture/release.

`CrossBuilder` is the type that process an entire os while `CrossBuilderCommand` represent each step. Maybe `CrossBuilderStep` would be better here.

internal/command/container.go
Bluebugs: Thinking about it, `ContainerRunner` is basically the engine definition (podman, docker, kubernetes).

internal/command/container.go
Bluebugs: A `ContainerImage` define the os/architecture and link to the proper engine logic to build for.

`AllContainerRunner` is the common logic between all engine.
`AllContainerImage` is the common property/logic between all image.

I was expecting to have to replace some of those when doing the kubernetes engine, but in fact they are completely generic.

internal/command/container.go
Bluebugs: I think that might be the right terminology as we might not always require some form of container/pod for a specific engine, but they will always do some cross compilation.

internal/command/command.go
andydotxyz: OK, I think I follow. So I'm not sure why `CrossBuilderStep` is iterating over `Image`, wouldn't the builder be responsible for iterations so the build step is just 1 step?

internal/command/container.go
andydotxyz: Ah, so `Engine` could work, remove the duplication that's tripping me up?
Could use `Create` instead of `New` so it's not looking like a constructor?
Also I realise here there is `ImageContainer` vs `ContainerImage` - which is desired?

internal/command/container.go
andydotxyz: It sounds like the `AllContainerRunner` could then be `baseEngine` (not-exported base type for common code) and then the `AllContainerImage` could be `baseContainer` or `baseImage` or whatever the right name  is and the implementations could embed that the same way?

internal/command/container.go
andydotxyz: Cool. also, as discussed above today, perhaps `Engine`?

Bluebugs:
Sounds like a lot of good suggestion. I can have a pass at it tomorrow and we can iterate on the change after that.

internal/command/command.go
Bluebugs: Indeed, the `CrossBuilder` will iterate over all the `ContainerImage` and call `CrossBuilderStep` once to do OS specific task.

internal/command/container.go
Bluebugs: I think `ContainerImage`.

internal/command/command.go
andydotxyz: So `RunEach` is more like just `Run`?

internal/command/container.go
Bluebugs: After trying to remove them, it actually is more useful than expected as code use a `ContainerImage` to build figure out what command to execute. A `ContainerImage` is a representation of an Engine/OS/Architecture capability. I think they are for the moment useful where they are.

internal/command/command.go
```diff
@@ -22,6 +22,69 @@ type Command interface {
 	Run() error                // Run runs the command
 }
 
+type PlatformSpecific interface {
+	Step(image ContainerImage) (string, string, error) // Called to build each possible architecture/OS combination
+}
+
+type CrossBuilder struct {
+	PlatformSpecific
```
andydotxyz, line 30: This is a bit confusing... really this should be a field I think - the CrossBuilder itself does not implement `PlatformSpecific` does it?

Perhaps `builder PlatformBuilder` works and `PlatformSpecific.Step` is named `PlatformBuilder.Build` perhaps? This way CrossBuilder delegates build to a platform specific element. If I'm right that this is what's happening I think the naming would flow.

andydotxyz (commented):
Thanks, this is looking good. One little comment about inheritance of interface where I think it should be just an inner field, and some name tweaks which I think would finish it up nicely.

andydotxyz (changes requested):
Thanks, this is looking good. One little comment about inheritance of interface where I think it should be just an inner field, and some name tweaks which I think would finish it up nicely.

Bluebugs:
Thanks @andydotxyz for the review. @lucor if you do have some time to look at this PR over the weekend, would really appreciate!

internal/command/container.go
```diff
@@ -0,0 +1,398 @@
...
+	gowindresBin = "/usr/local/bin/gowindres"
+)
+
+type Debugger interface {
+	Debug(v ...interface{})
+	Debugging() bool
+}
+
+type ContainerEngine interface {
+	CreateContainerImage(arch Architecture, OS string, image string) ContainerImage
```
lucor, line 28: It looks like `CreateContainerImage` could be moved as package function having it to accept a `ContainerEngine` argument. In this way we could remove it from the interface. Something like:
```CreateContainerImage(ce ContainerEngine, arch Architecture, OS string, image string) ContainerImage```

**Comment by andydotxyz** [comment]
Looks like the new code requires Go 1.16 but the project still supports 1.15 so the builds are failing.

File: internal/command/container.go (hunk 10 above)
Comment (Bluebugs) [reply internal/command/container.go:28]: I tried to figure how to make this work, but CreateContainerImage is allowing each engine to customize its behavior for a ContainerImage. I don't see how to do the dispatching properly without having it in the interface or at least a version of it in the interface, do you have any suggestion?

Comment (lucor) [reply internal/command/container.go:28]: The idea was to have it as a package function.
Could that work?

Comment (Bluebugs) [reply internal/command/container.go:28]: Trying to think about it, but the idea would be to move the function and the engine into a package for that specific engine? It would add an additional dispatching function that every OS specific file would use to route the call between different engine. Seems not ideal to me, but doable. Do I understand what you have in mind correctly?

**Comment by Bluebugs** [comment]
> Looks like the new code requires Go 1.16 but the project still supports 1.15 so the builds are failing.

It was actually the tool we use for the github action that need 1.18, not the code introduced in this PR. All green now.

File: internal/command/container.go [hunk 11]
```diff
@@ -0,0 +1,398 @@
...
+	Architecture() Architecture
+	OS() string
+	ID() string
+	Target() string
+	Env(string) (string, bool)
+	SetEnv(string, string)
+	SetMount(string, string)
+	AppendTag(string)
+
+	Engine() ContainerEngine
```
andydotxyz, line 60: It seems strange that an `Image` has an `Engine` - is that the right way round for the relationship?

andydotxyz (changes requested):
This is getting close - but there is still a confusing relationship between `CrossBuilder` and `PlatformBuilder`. The `Android` type seems to be both, but one requires the other...

And Images have Engines, which seems backwards somehow... I assumed an image was a simple static bundle and an engine ran it?

andydotxyz:
This is getting close but I am still lost in the terminology. Relationship between platform and cross builders is strange (android is both) and also image and engine are not in the dependency order I expected.
Is it possible to have a simple diagram or text overview of how these types are expected to fit together?

internal/command/container.go
```diff
@@ -0,0 +1,398 @@
...
+)
+
+type Debugger interface {
+	Debug(v ...interface{})
+	Debugging() bool
+}
+
+type ContainerEngine interface {
+	CreateContainerImage(arch Architecture, OS string, image string) ContainerImage
+	Debugger
```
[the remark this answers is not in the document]
Bluebugs: There is a few place where I rely on the ability to do Engine().Debugging(). Without that interface here, I would have to add another entry point on the container image or the engine to get to the debugger interface.

internal/command/container.go
Bluebugs: It is necessary in many part of the code to find the Engine related to an Image being processed. Storing that relationship simplify the code greatly. Only relying on the Engine->Image relationship would likely result in adding the engine to every function call everywhere and reduce readability.

internal/command/android.go
```diff
@@ -24,17 +24,16 @@ var (
...
-}
+var _ PlatformBuilder = (*Android)(nil)
+var _ Command = (*Android)(nil)
 
-// Description returns the command description
-func (cmd *Android) Description() string {
-	return "Build and package a fyne application for the android OS"
+func NewAndroidCommand() *Android {
+	r := &Android{CrossBuilder: CrossBuilder{name: "android", description: "Build and package a fyne application for the android OS"}}
+	r.builder = r
```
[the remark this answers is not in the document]
Bluebugs: The `CrossBuilder` is where the common code is and it is the interface called by the upper command. Still that `CrossBuilder` need to know details of what it is building, which is what the `PlatformBuilder` do. Provide the `CrossBuilder` logic the ability to build for a specific platform.

The other source of the design is that we are not using cli.App at the moment, but a `Command` structure. I think that there will be further reorganization here once we move to cli.App as part of preparing the move to fyne repository. Something I would like to push back to a later PR and so I have avoided so far touching much of the `Command` design for that reason.

internal/command/command.go
```diff
@@ -22,6 +22,69 @@ type Command interface {
...
 }
 
+type PlatformBuilder interface {
+	Build(image ContainerImage) (string, string, error) // Called to build each possible architecture/OS combination
+}
+
+type CrossBuilder struct {
+	builder PlatformBuilder
+
+	Images []ContainerImage
```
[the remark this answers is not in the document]
Bluebugs: Indeed, forgot that one.

internal/command/command.go
```diff
@@ -22,6 +22,69 @@ type Command interface {
 	Run() error                // Run runs the command
 }
 
+type PlatformBuilder interface {
```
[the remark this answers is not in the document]
Bluebugs: It is a good question. If we want to allow outside tool to directly use fyne-cross code, yes, otherwise, no. We can keep them private at the moment, and when we move things to fyne we can decide on a struct by struct basis.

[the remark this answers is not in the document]
Bluebugs: I have unexported all the new type and functions. I have kept the exported type for `Command` as that should be addressed when migrating to cli.App.

Bluebugs:
Small attempt at a diagram, not sure it is super usable:

```mermaid
classDiagram
   Command<|--Linux
   Command<|--Windows
   CrossBuilder<|--Linux
   CrossBuilder<|--Windows

   class Command{
     <<Interface>>
     +Run() error
   }
   class CrossBuilder{
     <<Interface>>
     +Build(ContainerImage)
     -setupContainerImage(flags, args) error
   }

   Command-->ContainerImage:Run() Use
   CrossBuilder-->ContainerImage:Build() Use
   CrossBuilder-->ContainerEngine:setupContainerImage() Use

   ContainerEngine<|--LocalContainerEngine
   ContainerImage<|--LocalContainerImage
   ContainerEngine<|--KubernetesContainerEngine
   ContainerImage<|--KubernetesContainerImage

   class ContainerEngine{
     <<Interface>>
     +createContainerImage(arch Architecture, OS string, image string) containerImage
   }

   class ContainerImage{
     <<Interface>>
     +Run(vol volume.Volume, opts options, cmdArgs []string) error
   }
```

andydotxyz:
That is really helpful thanks - thought it is missing the PlatformBuilder which was part of my confusion from earlier.
Apart from that the diagram makes sense - if it included the Fields that were of the object types as well we might see the circular relationships emerge.

internal/command/container.go
andydotxyz: I can see why it might need it internally - but why does it have to be exposed through the interface?

internal/command/container.go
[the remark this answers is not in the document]
andydotxyz: But the ability to debug is not unique to a `ContainerEngine` so should it not be possible to type check and call debug if it is there? A helper method could do that for you if it's repetitive I would think.
Either way debugging is not part of the behaviour required, which is why separating them was suggested in the first place - this version is not separated, it's just moved.

internal/command/android.go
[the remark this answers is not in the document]
andydotxyz: It seems like this `CrossBuilder` is a helper utility then, so it could be an internal detail and not part of the interface heirarchy. It could be a field within the platform builder and created as needed where the platform builder could be passed in - this removes the circular nature of the relationship.

Bluebugs:
Adding `platformBuilder` which is being used by `CrossBuilder` to provide per platform specific build step (The OS specific structure also rely on `CrossBuilder` to provide actually a portion of the Command interface that was duplicated across all OS before). I think the arrow for most field that have structure or interface are actually visible here.

```mermaid
classDiagram
   Command<|--Linux
   Command<|--Windows
   platformBuilder<|--Linux
   platformBuilder<|--Windows

   class Command{
     <<Interface>>
     +Run() error
   }
   class CrossBuilder{
     <<Interface>>
     +platformBuilder
     +Run()
     -setupContainerImage(flags, args) error
   }
   class platformBuilder{
     <<Interface>>
     +Build(ContainerImage)
   }

   CrossBuilder-->ContainerImage
   CrossBuilder-->ContainerEngine:setupContainerImage() Use
   CrossBuilder-->platformBuilder:Build() Use
   platformBuilder-->ContainerImage
   Command<|--CrossBuilder

   ContainerEngine<|--LocalContainerEngine
   ContainerImage<|--LocalContainerImage
   ContainerEngine<|--KubernetesContainerEngine
   ContainerImage<|--KubernetesContainerImage

   class ContainerEngine{
     <<Interface>>
     +createContainerImage(arch Architecture, OS string, image string) containerImage
   }

   class ContainerImage{
     <<Interface>>
     +Run(vol volume.Volume, opts options, cmdArgs []string) error
   }

   ContainerEngine-->ContainerImage:create
```

internal/command/container.go
```diff
@@ -59,6 +58,7 @@ type containerImage interface {
 	Tags() []string
 
 	Engine() containerEngine
+	Debugger() debugger
```
andydotxyz, line 61: This looks more like moved rather than removed. In this new position all containerImages must be debuggable. Is this a desirable definition of an image?
Go loose typing indicates that anything could be debuggable, if it provides the facilities to be debugged, rather than stating which interfaces will have to support debugging...

andydotxyz:
> Adding `platformBuilder` which is being used by `CrossBuilder` to provide per platform specific build step

Thanks for this, I think I  am starting to understand.
Proposal: `CrossBuilder is not an interface, it is our single struct type that is the core of this package. It would use a `PlatformBuilder` and `ContainerEngine` to function.

> The OS specific structure also rely on CrossBuilder to provide actually a portion of the Command interface that was duplicated across all OS before

This feels messy - combining re-use with inheritance. Why not separate the requirement on `CrossBuilder` with some "command common code" type (I don't know what scope this has) that contains only this common code? Remember that Go supports multiple inheritance. This way the platform builders don't depend on the cross builder, which depends on them again.

I hope these ideas are helpful, It should take out some of the seemingly circular usages.

Bluebugs:
> > Adding `platformBuilder` which is being used by `CrossBuilder` to provide per platform specific build step
> 
> Thanks for this, I think I am starting to understand. Proposal: `CrossBuilder is not an interface, it is our single struct type that is the core of this package. It would use a `PlatformBuilder`and`ContainerEngine` to function.

It is pretty much this already, except it use a ContainerImage to function which is provided by a ContainerEngine.

> > The OS specific structure also rely on CrossBuilder to provide actually a portion of the Command interface that was duplicated across all OS before
> 
> This feels messy - combining re-use with inheritance. Why not separate the requirement on `CrossBuilder` with some "command common code" type (I don't know what scope this has) that contains only this common code? Remember that Go supports multiple inheritance. This way the platform builders don't depend on the cross builder, which depends on them again.

I kind of want to avoid touching much of the command part as it will disappear, I believe, once we migrate to cli.App.

internal/command/container.go
Bluebugs: Hum, I am not sure that I get what you want correctly. Do you mean, I should always cast all Debugger call on the Engine()? Something like: `debug, ok := image.Engine().(Debugger)` and then if the debugger is present do call the debugger function?

andydotxyz:
> > > The OS specific structure also rely on CrossBuilder to provide actually a portion of the Command interface that was duplicated across all OS before
> > 
> > 
> > This feels messy - combining re-use with inheritance. Why not separate the requirement on `CrossBuilder` with some "command common code" type (I don't know what scope this has) that contains only this common code? Remember that Go supports multiple inheritance. This way the platform builders don't depend on the cross builder, which depends on them again.
> 
> I kind of want to avoid touching much of the command part as it will disappear, I believe, once we migrate to cli.App.

I see - but how does the current cli library require that each platform-specific builder rely on the cross builder?

Bluebugs:
> I see - but how does the current cli library require that each platform-specific builder rely on the cross builder?

The cross builder is the provider of Run() which really shouldn't be duplicated, but due to the Cli library requiring the existence of a Run per platform it has to be somewhere. The solution is to either duplicate code (bad) or reuse by depending on cross builder (better). Best will be when we do not need to link the crossBuilder.Run() with the Cli and can then just drop that from the dependencies.

andydotxyz:
After discussing debugger on the call this is closer but I am still stuck on :

> The solution is to either duplicate code (bad) or reuse by depending on cross builder (better).

isn’t there a third option where they don’t have to depend on one another but the common code is used by both?

Bluebugs:
> > The solution is to either duplicate code (bad) or reuse by depending on cross builder (better).
> 
> isn’t there a third option where they don’t have to depend on one another but the common code is used by both?

I can not think of a different solution until we use cli.App that would be significantly different from any of the one above. I am convinced that using cli.App will solve this problem and that we won't need anymore that duplicated code, but switching to cli.App introduce massive change in the code base and this PR is already big enough that I do not want to do that here. Still I think that that would likely address your problem completely and be your third option.

andydotxyz:
Please don't force push after reviews started, it now needs a complete review again as we can't see the changes any more

Bluebugs:
> Please don't force push after reviews started, it now needs a complete review again as we can't see the changes any more

Sorry about that. I made a mistake in merging the wrong branch and I had to force push to undo that change.

andydotxyz:
I had a look at the code and am not sure I agree that crossBuilder must exist because of the way the CLI works.
I see that it removes duplication of shared code, but I don't think it's the right way to do it because `crossBuilder` and `platformBuilder` create a tight circular dependency.

Looking at it there are 3 methods, two if which just return the name and description, which is different. The third, `Run` does all of the hard work. So why not:
* return Name and Description back to the commands (platform builder entry points)
* move the `crossBuilder.Run` to a helper method and remove `crossBuilder` completely
* call the helper run method from each platform builder `Run`

This should be approx the same number of lines as you have, but with a cleaner control dependency graph.

andydotxyz:
> The build failure is unrelated as far as I know and due to a bug with v2.2

Or maybe it's not a bug and this is now using Fyne v2.2.1 tools to compile a Fyne v2.1.4 app - we do ask that folk keep the versions in sync.

Probably we didn't see this coming but it is a side/effect of using the generated code to insert metadata. I guess we could improve the tooling to try and detect if `SetMetadata` is available.

andydotxyz:
I have updated the calculator module and kicked the tests so we can see if that was the only error.

andydotxyz (approved):
I'm happy with how this turned out. I'd like to know that @lucor is too though :)

andydotxyz:
Perhaps we should make the current release, then land it to work on with more time?

Issue #97, opened 2022-03-29
Windows builds no longer pass "-H windowsgui"


Jacalz:
### Describe the bug:

Windows binaries seems to no longer pass "-H windowsgui", which means that all binaries (even those built without `-console`) show a console when opened. This was recently reported to me at https://github.com/Jacalz/rymdport/issues/34.

I suspect that this broke with https://github.com/fyne-io/fyne-cross/pull/42/files#r692376503 but I sadly don't have time to look at it right now.

### To Reproduce:
Steps to reproduce the behaviour:
1. Build a binary for windows without the `-console` flag.
2. Run the binary on Windows and notice that console is shown.

### Example code:
Any code will do.

### Device and debug info (please complete the following information):

<summary>Device info</summary>

 - **OS:**  Fedora Workstation
 - **Version:**  35 with kernel 5.16.17
 - **Go version:**  1.16.15
 - **fyne-cross version:**  1.2.0
 - **Fyne version:**  2.1.3 (not relevant)

<details><summary>Debug info</summary><br><pre>
I did not bother with --no-cache because `--debug` has enough meaningful information.
```bash
[jacob@vostro rymdport]$ ~/go/bin/fyne-cross windows -arch amd64 --debug
[i] Target: windows/amd64
command.Context{Volume:volume.Volume{binDirHost:"/home/jacob/Git/rymdport/fyne-cross/bin", cacheDirHost:"/home/jacob/.cache/fyne-cross", distDirHost:"/home/jacob/Git/rymdport/fyne-cross/dist", tmpDirHost:"/home/jacob/Git/rymdport/fyne-cross/tmp", workDirHost:"/home/jacob/Git/rymdport"}, Architecture:"amd64", Env:[]string{"GOOS=windows", "GOARCH=amd64", "CC=x86_64-w64-mingw32-gcc"}, ID:"windows-amd64", LdFlags:[]string{"-H=windowsgui"}, OS:"windows", Tags:[]string(nil), AppBuild:"16", AppID:"io.github.jacalz.rymdport", AppVersion:"3.0.0", CacheEnabled:true, DockerImage:"docker.io/fyneio/fyne-cross:1.2-windows", Icon:"internal/assets/icon/icon-512.png", Name:"rymdport.exe", Package:".", Release:false, StripDebug:true, Debug:true, Pull:false, Category:"", Certificate:"", Developer:"", Keystore:"", KeystorePass:"", KeyPass:"", Password:"", Profile:""}
[...]

Issue #96, opened 2022-03-28
Bug in -tag flag


luisaodh:
### Describe the bug:

While compiling, I came across a small discrepancy with the -tag flag.
For compiling I am using the following command:
```fyne-cross linux -arch arm -tags wayland,gles,mesa```
To be able to use this command, I had to use fyne-cross version 1.1.0, because the usage of the version 1.2.0 leads in the error:

[i] Target: linux/arm
[i] Cleaning target directories...
[✓] "bin" dir cleaned: /home/devel/Dokumente/smart-home/fyne-cross/bin/linux-arm
[✓] "dist" dir cleaned: /home/devel/Dokumente/smart-home/fyne-cross/dist/linux-arm
[✓] "temp" dir cleaned: /home/devel/Dokumente/smart-home/fyne-cross/tmp/linux-arm
[i] Checking for go.mod: /home/devel/Dokumente/smart-home/go.mod
[✓] go.mod found
[i] Building binary...
cmd/go: -tags space-separated list contains comma
[✗] exit status 2

But when I display the help menu in the command line (```fyne-cross linux -arch arm --help```) it says for the desription of the -tags: ```List of additional build tags separated by comma.```
Is this meant to be like this or a bug? And how am I supposed to execute my command in version 1.2.0?

lucor:
Fix available for testing in https://github.com/fyne-io/fyne-cross/pull/99

Jacalz:
This will be fixed with the next release. You can test it on the `develop` branch right now if you want.

PR #103, opened 2022-04-09
Release 1.2.1

lucor:
### Added

- Added the `--engine` flags that allows to specify the container engine to use
  between docker and podman. The default behavior is not changed, if the flag is
  not specified fyne-cross will auto detect the engine.

### Fixed 

- Windows builds no longer pass "-H windowsgui" #97
- Multiple tags cannot be specified using the `-tags` flag #96


Jacalz (approved):
This looks good to me. The tags feature (that I missed testing last time) also seems to be working perfectly on my end.

We did talk about releasing a v1.3.0 with Go 1.18.1 and Zig, but you have my approval here as well if you wish to release v1.2.1 before that. Do as you wish :)

lucor:
Let's release 1.2.1 since fixes some regressions :)

PR #108, opened 2022-04-13
Feature/kubernetes

### LINKED ISSUE #98: Add Kubernetes cluster support

Bluebugs:
### Description:
This add a working Kubernetes support for fyne-cross. It does use S3 to upload/download information out of the container/pod. The pod run for ever using a sleep infinite (I think we could hard code a limit there of maybe 30min?). Kubernetes exec command are used to execute command inside the pod. For most case, we do not need to change directory, but when we do I have to add a shell command to solve this problem.

I have moved a few operation that where done on the host, but relied on the file being local to be inside the pod (zip, move, copy). I haven't figured out how to handle the case when no Icon is provided at the moment.

I still need to do extensive testing with this patch (with kubernetes and without), but I wanted to start getting people review as early as possible.

Fixes #98 

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


Bluebugs:
@metal3d I couldn't add you as a reviewer, but if you do have some time have a look.

Bluebugs:
> We can't use `io/fs` yet, as we support down to Go 1.14 for our developers.

Oh, I was wondering what was that error. Thanks, will fix.

Bluebugs:
Closing this PR as the amount of change after the refactoring PR land is going to be really massive and make this harder to understand. Will do a clean PR on top of the refactoring PR.

PR #112, opened 2022-04-18
Add support for .xip file format of XCode download

andydotxyz:
Keep dmg support for older files

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


Bluebugs (approved):

andydotxyz:
This PR should not be accepted as is - the xip file is XCode not tools so build would have to invoke other osxcross scripts. Best to close it I think and if required add support to extract from XCode download later. Seems unlikely as the file is huge and not something developers would keep around.

PR #115, opened 2022-05-19
Update go and fyne versions

russellsteadman:
### Description:

Updates Docker base to Go v1.18.2, Fyne v2.1.4

### Checklist:

Note: `develop` branch was already testing on Go v1.18.X in `ci.yml`

- [x] Tests included.
- [x] Tests all pass.


Jacalz:
Thanks for the PR. I think we were talking about getting Go 1.18 rolled out together with some other larger work but I’m not sure. Cc @lucor.

lucor:
Yes, the plan was to release a new version of fyne-cross with Go 1.18.x when fyne 2.2.0 is out.

andydotxyz:
It is time :) Plus an update with some important fixes followed, so this should now use Fyne `v2.2.1`

docker/base/Dockerfile
```diff
@@ -1,6 +1,6 @@
-ARG GO_VERSION=1.17.8
+ARG GO_VERSION=1.18.2
 # fyne stable branch
-ARG FYNE_VERSION=v2.1.3
+ARG FYNE_VERSION=v2.1.4
```
lucor, line 3:
```suggestion
ARG FYNE_VERSION=v2.2.1
```

lucor (changes requested):
As @andydotxyz pointed out into the comment the new version of Fyne is out :)
If you have any chance to update your PR using the latest versions of Fyne CLI and Go, we should be ready to go

russellsteadman:
@lucor Updated per suggestions

lucor (approved):
Thanks

PR #127, opened 2022-08-23
Feature/strip

### LINKED ISSUE #126: Missing application icon when using -release in windows
### Describe the bug:
When I build with `fyne-cross windows -release --no-cache` everything is fine and it compile sucessfully, but if I add the `-release` flag error message `Missing application icon at "C:\Users\DELL\GolandProjects\PotionCraftSaveEditor\fyne-cross\tmp/windows-amd64/Icon.png"` appears and cross stopped.

On that tmp directory, only a file named `PotionCraftSaveEditor.exe.ico` exist.

### To Reproduce:
Build with `fyne-cross windows -release --no-cache`

### Example code:
https://github.com/foxwhite25/PCSE

### Device and debug info (please complete the following information):

<summary>Device info</summary>

 - **OS:** Windows 10
 - **Go version:** 1.18
 - **fyne-cross version:** 1.3
 - **Fyne version:** 2.2.3

<details><summary>Debug info</summary><br><pre>
PS C:\Users\DELL\GolandProjects\PotionCraftSaveEditor> ./fyne-cross.exe windows -release -arch amd64,386 --no-cache --debug
[i] Target: windows/amd64
command.Context{Volume:volume.Volume{binDirHost:"C:\\Users\\DELL\\GolandProjects\\PotionCraftSaveEditor\\fyne-cross\\bin", cacheDirHost:"C:\\Users\\DELL\\AppData\\Local\\fyne-cross", distDirHost:"C:\\Users\\DELL\\GolandProjects\\PotionCraftSaveEditor\\fyne-cross\\dist", tmpDirHost:"C:\\Users\\DELL\\GolandProjects\\PotionCraftSaveEditor\\fyne-cross\\tmp", workDirHost:"C:\\Users\\DELL\\GolandProjects\\PotionCraftSaveEditor"}, Architecture:"amd64", Engine:command.Engine{Name:"docker", Binary:"C:\\Program Files\\Docker\\Docker\\resources\\bin\\docker.exe"}, Env:map[string]string{"CC":"x86_64-w64-mingw32-gcc", "GOARCH":"amd64", "GOOS":"windows"}, ID:"windows-amd64", LdFlags:[]string{"-H=windowsgui"}, OS:"windows", Tags:[]string(nil), AppBuild:"1", AppID:"", AppVersion:"1.0", CacheEnabled:false, DockerImage:"docker.io/fyneio/fyne-cross:1.3-windows", Icon:"Icon.png", Name:"PotionCraftSaveEditor.exe", Package:".", Release:true, StripDebug:true, Debug:true, Pull:false, Category:"", Certificate:"", Developer:"", Keystore:"", KeystorePass:"", KeyPass:"", Password:"", Profile:""}
[✓] "bin" dir cleaned: C:\Users\DELL\GolandProjects\PotionCraftSaveEditor\fyne-cross\bin\windows-amd64
[✓] "dist" dir cleaned: C:\Users\DELL\GolandProjects\PotionCraftSaveEditor\fyne-cross\dist\windows-amd64
[✓] "temp" dir cleaned: C:\Users\DELL\GolandProjects\PotionCraftSaveEditor\fyne-cross\tmp\windows-amd64
[i] Checking for go.mod: C:\Users\DELL\GolandProjects\PotionCraftSaveEditor\go.mod
[✓] go.mod found
fyne cli version: v2.2.3
C:\Users\DELL\go\bin\fyne.exe release -os windows -name PotionCraftSaveEditor.exe -icon C:\Users\DELL\GolandProjects\PotionCraftSaveEditor\fyne-cross\tmp/windows-amd64/Icon.png -appBuild 1 -appVersion 1.0
Missing application icon at "C:\Users\DELL\GolandProjects\PotionCraftSaveEditor\fyne-cross\tmp/windows-amd64/Icon.png"
[✗] could not package the Fyne app: could not package the Fyne app: exit status 1
</pre></details>

Bluebugs:
### Description:
This enable -no-strip-debug for platform that we were actually not triggering it for (android, ios and windows) and make them equivalent as what we are doing for linux, darwin and freebsd.

Fixes #126 

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/internal/command/container.go
+++ b/internal/command/container.go
@@ -219,19 +219,18 @@ func goBuild(ctx Context, image containerImage) error {
 	return nil
 }
 
-// fynePackage package the application using the fyne cli tool
-func fynePackage(ctx Context, image containerImage) error {
+func fyneCommand(command string, ctx Context, image containerImage) ([]string, error) {
 	if debugging() {
 		err := image.Run(ctx.Volume, options{}, []string{fyneBin, "version"})
 		if err != nil {
-			return fmt.Errorf("could not get fyne cli %s version: %v", fyneBin, err)
+			return nil, fmt.Errorf("could not get fyne cli %s version: %v", fyneBin, err)
 		}
 	}
 
 	target := image.Target()
 
 	args := []string{
-		fyneBin, "package",
+		fyneBin, command,
 		"-os", target,
 		"-name", ctx.Name,
 		"-icon", volume.JoinPathContainer(ctx.TmpDirContainer(), image.ID(), icon.Default),
@@ -248,6 +247,16 @@ func fynePackage(ctx Context, image containerImage) error {
 	tags := image.Tags()
 	if len(tags) > 0 {
 		args = append(args, "-tags", fmt.Sprintf("%q", strings.Join(tags, ",")))
+	}
+
+	return args, nil
+}
+
+// fynePackage package the application using the fyne cli tool
+func fynePackage(ctx Context, image containerImage) error {
+	args, err := fyneCommand("package", ctx, image)
+	if err != nil {
+		return err
 	}
 
 	// Enable release mode, if specified
@@ -273,7 +282,7 @@ func fynePackage(ctx Context, image containerImage) error {
 		WorkDir: workDir,
 	}
 
-	err := image.Run(ctx.Volume, runOpts, args)
+	err = image.Run(ctx.Volume, runOpts, args)
 	if err != nil {
 		return fmt.Errorf("could not package the Fyne app: %v", err)
 	}
@@ -283,34 +292,9 @@ func fynePackage(ctx Context, image containerImage) error {
 // fyneRelease package and release the application using the fyne cli tool
 // Note: at the moment this is used only for the android builds
 func fyneRelease(ctx Context, image containerImage) error {
-	if debugging() {
-		err := image.Run(ctx.Volume, options{}, []string{fyneBin, "version"})
-		if err != nil {
-			return fmt.Errorf("could not get fyne cli %s version: %v", fyneBin, err)
-		}
-		return nil
-	}
-
-	target := image.Target()
-
-	args := []string{
-		fyneBin, "release",
-		"-os", target,
-		"-name", ctx.Name,
-		"-icon", volume.JoinPathContainer(ctx.TmpDirContainer(), image.ID(), icon.Default),
-		"-appBuild", ctx.AppBuild,
-		"-appVersion", ctx.AppVersion,
-	}
-
-	// add appID to command, if any
-	if ctx.AppID != "" {
-		args = append(args, "-appID", ctx.AppID)
-	}
-
-	// add tags to command, if any
-	tags := image.Tags()
-	if len(tags) > 0 {
-		args = append(args, "-tags", fmt.Sprintf("%q", strings.Join(tags, ",")))
+	args, err := fyneCommand("release", ctx, image)
+	if err != nil {
+		return err
 	}
 
 	// workDir default value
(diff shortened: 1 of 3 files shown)


andydotxyz (commented):
I like this change, but do we need to document the behaviour change?
I think this will add `-release` parameter by default for builds. What do @lucor and @Jacalz think about this?

andydotxyz:
Also, it seems to break Android builds. Any idea what happened?

andydotxyz (changes requested):
Ah I figured out the Android issue. Looks like this change breaks release builds as the `-release` flag is no longer respected. This seems to only be used for Android testing (see the note in container.go "// Note: at the moment this is used only for the android builds").

Checking `ctx.Release` needs to be re-added and the tests should start working again

internal/command/container.go
```diff
@@ -250,9 +249,14 @@ func fynePackage(ctx Context, image containerImage) error {
 		args = append(args, "-tags", fmt.Sprintf("%q", strings.Join(tags, ",")))
 	}
 
-	// Enable release mode, if specified
-	if ctx.Release {
-		args = append(args, "-release")
+	return args, nil
+}
+
+// fynePackage package the application using the fyne cli tool
```
Jacalz, line 255:
```suggestion
// fynePackage packages the application using the fyne cli tool
```

Jacalz (changes requested):
I found one suggestion for a comment. However, I am not quite sure what this is trying to do. Where is `ctx.StripDebug` set and why is it only checked if it isn't targeting Linux, macOS or FreeBSD?

Bluebugs:
> I found one suggestion for a comment. However, I am not quite sure what this is trying to do. Where is `ctx.StripDebug` set and why is it only checked if it isn't targeting Linux, macOS or FreeBSD?

Other platform do not use fyne to do the build. When the build are directly done by fyne-cross it does honor the StripDebug (which is the result of a flag set on the command line). The problem was that before this, the platform build by fyne would not get stripped.

Jacalz:
Ah, I see. That makes sense but I think you might want to add a comment about that in the code. It is a bit confusing.

Bluebugs:
The change in android image are not propagated to the test and so test will still fail. If you recompile locally the image so you can have bundletool in it, everything will work as one expect.

lucor (dismissed):
LGTM. 
As discussed on the slack channel, let's have the android image updated and this tested manually as part of #134.

Before merge, do you mind to fix the small typo into the doc comment spotted by @Jacalz during the review?

Bluebugs:
Oh indeed, I tried to commit @Jacalz change from github and it failed due to some wrong email somewhere and forgot after that. Will adress.

lucor (approved):

andydotxyz:
The failed Android builds is concerning me.
It seems that the interpretation of the release has changed and now the build is attempting to package for google play instead of verifying the build process.
Adding the `bundletool` may have been an attempt to fix this, but is it the right fix? Are we attempting to do a store build in these build checks?

Bluebugs:
We should be able to move this forward with #140

lucor:
Merging is blocked since the android tests still fail.
They should pass once #141 is merged and the new docker images are available.

Bluebugs:
@Jacalz @andydotxyz with the new image with latest fyne and go, we are now all green here!

andydotxyz (approved):
Yup, Sorry I was expecting a rebase or something but it seems green now

Bluebugs:
@Jacalz would you mind checking if this PR now match your requested change earlier?

Jacalz (approved):
Sure. Looks good to me

PR #143, opened 2022-11-16
Normalise executable name for darwin

### LINKED ISSUE #139: fyne-cross doesn't normalize executable name
### Describe the bug:
fyne package does normalize the executable name to fit in filesystem better, while fyne-cross does use the application name. This lead to inconsistency when building for darwin on darwin (as fyne-cross use fyne package in that case) and might likely cause bugs in some case.

For reference fyne package normalization function is here: https://github.com/fyne-io/fyne/blob/42fa9008fad15ae2ba5b52182865cfb18647fec3/cmd/fyne/internal/commands/package.go#L389

### To Reproduce:
Steps to reproduce the behaviour:
1. Create an app that has a name which does not match the module name.
2. Use fyne-cross to generate a package for darwin on darwin
3. Use fyne-cross to generate a package for darwin from another OS
4. Binary names differ inside the package

itsjustdel:
This fix replicates how `fyne package` names the executable for darwin, resulting in the binary being named after the module name which is read from `go.mod` 

An example of an app named "Nomad Toml Name" (set in the `toml`) from a module named "nomad"

image

 

Fixes #139

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/command/container.go
+++ b/internal/command/container.go
@@ -315,7 +315,7 @@ func calculateExeName(sourceDir, os string) string {
 		}
 	}
 
-	if os == "windows" {
+	if os == windowsOS {
 		exeName = exeName + ".exe"
 	}
 
--- a/internal/command/container.go
+++ b/internal/command/container.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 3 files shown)


internal/command/container.go
```diff
@@ -285,6 +299,28 @@ func fynePackage(ctx Context, image containerImage) error {
 	return nil
 }
 
+func calculateExeName(sourceDir, os string) string {
```
lucor, line 302: What about adding a doc comment explaining that it is ported from the fyne base code and why?
It'd useful for reference keeping track of the reason we have some logic regarding windows even if has impact only for darwin.

lucor (commented):
Thanks @itsjustdel, some minor comments inline

internal/command/container.go
itsjustdel: Great idea

andydotxyz (approved):

lucor (approved):
Thanks @itsjustdel

PR #145, opened 2022-11-23
Bugs/propage architecture for local build

Bluebugs:
### Description:
Fix local build when cross compiling. In the case where you for example use a darwin arm64 host and are cross compiling for darwin amd64, the local build path was not propagating the architecture information correctly.

### Checklist:
- [ ] Tests all pass.


lucor (approved):
Good catch

Issue #107, opened 2022-04-11
Integrate fyne-cross and fyne CLI


Bluebugs:
### Is your feature request related to a problem? Please describe:
fyne-cross need to follow fyne behavior closely and having two different repository increase complexity in managing two different release cycle. Integrating fyne-cross into fyne, could also lead to better developers experience by probing the system and falling back to fyne-cross when cross compiling. It could also improve functionality discoverability by having only one command and tool to install.

### Is it possible to construct a solution with the existing API?
It is, by calling fyne-cross and fyne independently, but it isn't giving the best possible experience.

### Describe the solution you'd like to see:
This issues is about figuring out a path forward to merge fyne-cross into fyne. Figuring out what we want to improve in fyne-cross code base prior to merging and then what are the step to do that merging.

Rough idea here:
1. Migrate fyne-cross to use cli.App
2. Move Docker image to an independent repository (would be nice to automate their build on release)
3. Address the darwin image specificity
4. Change fyne-cross path to match fyne CLI path
5. Import fyne-cross git into fyne CLI
6. Plug fyne cross command in
7. Add logic to automatically detect if the system should fallback to fyne cross
8. Release fyne 2.3
9. Archive fyne-cross git repository
10. Enjoy :-)

Let me know what you think and if that path make sense to you.

PR #148, opened 2022-12-07
docker: update fyne-cross to support images from the fyne-cross-images repo

lucor:
### Description:

This PR updates fyne-cross to use the images from the [fyne-cross-images](https://github.com/fyne-io/fyne-cross-images) repo.

Relates #107

### Checklist:

- [ ] Tests included.
- [X] Lint and formatter run with no errors.
- [X] Tests all pass except the build app tests that requires the new images published on DockerHub


Jacalz:
Should this also remove the images that are located here?

lucor:
> Should this also remove the images that are located here?



> Should this also remove the images that are located here?

Yes, this will remove all the images. They will be moved into the fyne-cross-images repo.

internal/command/windows.go
```diff
@@ -223,14 +236,115 @@ func (cmd *windows) setupContainerImages(flags *windowsFlags, args []string) err
...
 
 		cmd.Images = append(cmd.Images, image)
 	}
 
 	return nil
 }
+
+// makeWindowsResource create a windows resource under the project root
+// that will be automatically linked by compliler during the build
+func makeWindowsResource(ctx Context, image containerImage) (string, error) {
```
Bluebugs, line 253: Would it be actually possible to have this part of the image as one of the fyne-cross tool? This would allow to run this in the container and have it working with k8s.

lucor: Maybe I'm missing something but this should not affect the previous implementation.
`makeWindowsResource` is replacing the `WindowsResource` from the common.go file.
The main difference it is that now the windows resource creation is done using a Go library instead of the `windres` CLI tool. 

Note: I still need to add the `internal/cmd/fyne-cross-s3` to the new base image.

Bluebugs: Indeed, this is not a regression as this path is not working with the k8s port. I am actually wondering if we were to use fyne build if that would solve this problem. I can look into switching to use fyne build.

internal/command/container.go
```diff
@@ -175,6 +173,10 @@ func goBuild(ctx Context, image containerImage) error {
 	// add go build command
 	args := []string{"go", "build", "-trimpath"}
 
+	if ctx.BuildMode != "" {
```
Bluebugs, line 176: This will need to be rebased on develop and be added to GOFLAGS instead as that's how you can pass flags to the go command when using fyne build.

lucor: Merged develop in 7c2f186

Bluebugs:
I think it is failing even if the image are now in docker hub, because we are not passing -pull to the command line.

lucor:
> I think it is failing even if the image are now in docker hub, because we are not passing -pull to the command line.

It looks like that for some reasons not all the images are available yet

```
docker pull fyneio/fyne-cross-images:linux
Error response from daemon: manifest for fyneio/fyne-cross-images:linux not found: manifest unknown: manifest unknown
```

Bluebugs:
> ```
> docker pull fyneio/fyne-cross-images:linux
> Error response from daemon: manifest for fyneio/fyne-cross-images:linux not found: manifest unknown: manifest unknown
> ```

Ok, the linux image and the android image are now in the repository.

Bluebugs:
Almost there! We have a last issue with the the appid passed for the android build.

```
- os: android
  args: -app-id calc.${GITHUB_SHA}
```

I think, if we just change it to the following it would work:
```
- os: android
  args: -app-id calc.sha${GITHUB_SHA}
```

Bluebugs (approved):
Boom, this looks great! Nice work!

lucor:
Good catch, until now we were lucky about the SHA hash for the ID :)

PR #151, opened 2022-12-19
ability to change docker registry

### LINKED ISSUE #150: Being able to set docker registry
When we have our own docker registry mirror and we want to use that instead of dockerhub, we have to set full image name
and there is lots of images that fyne-cross is using so it's good to be able to set dockerhub mirror

Also the bad thing is tag `docker.io` is hardcoded

mhkarimi1383:
### Description:

New cmd arg for docker registry

Fixes #150 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

- [x] Any breaking changes have a deprecation path or have been discussed.
- [x] Updated the vendor folder (using `go mod vendor`).


mhkarimi1383:
My changes are not based of `develop` branch :(

PR #152, opened 2022-12-19
ability to change docker registry

### LINKED ISSUE #150: Being able to set docker registry

mhkarimi1383:
### Description:

Added new flag to handle docker registry override

Fixes #150 

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

- [x] Any breaking changes have a deprecation path or have been discussed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/ulikunitz/xz v0.5.10 // indirect
 	github.com/urfave/cli/v2 v2.11.1
-	golang.org/x/mod v0.7.0
+	golang.org/x/mod v0.3.0
 	golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f
 )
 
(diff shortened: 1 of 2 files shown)


mhkarimi1383:
Unit tests are getting error where I did nothing

andydotxyz:
The test failures I see (in fact they are compile errors) seem quite correct to me. In many places you have changed type of fields but I can't see why, this is likely the cause.

mhkarimi1383:
Oh
I will check them out
But some of them are from my editor linter

mhkarimi1383:
I reverted the changes in some test files
it should be okay now

mhkarimi1383:
the problem with that error is because of http2/transport.go updates, we have to upgrade the GoLang version or Downgrade related package

If we don't want to upgrade golang version I will downgrade

Bluebugs:
We are not planning to upgrade golang version requirement at this point indeed, so please downgrade for the moment.

mhkarimi1383:
> We are not planning to upgrade golang version requirement at this point indeed, so please downgrade for the moment.

I have downgraded the package now
If it fails I will revert the whole go.sum file

Bluebugs (approved):

PR #154, opened 2022-12-27
Follow rfc1034 like Fyne for package name.

Bluebugs:
### Description:
fyne-cross is not aware that fyne follow rfc1034 when packaging the application for ios. This PR correct that.

### Checklist:
- [ ] Tests all pass.


lucor (commented):
I'd only add some tests for the output of the rfcLabel func, especially for the edge cases with `.` and Unicode chars. Apart that LGTM

Bluebugs:
This won't be necessary after merging #159 .

PR #158, opened 2022-12-30
Allow to set up the executable name from the command line.

Bluebugs:
### Description:

This allow to mimic the executable name picked by fyne build command. I think this is another issue with not using fyne directly for building. I will be investigating switching more platform to build with fyne directly as a permanent fix for this serie of problem.

### Checklist:
- [ ] Tests all pass.

### DIFF
--- a/internal/command/container.go
+++ b/internal/command/container.go
@@ -1,6 +1,7 @@
 package command
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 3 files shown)


internal/command/container.go
```diff
@@ -205,8 +206,15 @@ func goBuild(ctx Context, image containerImage) error {
...
 
 	// set output folder to fyne-cross/bin/<target>
-	binaryName := ctx.Name
+	binaryName := ctx.ExecutableName
+	if binaryName == "" {
+		binaryName = ctx.Name
+	}
 	if image.OS() == darwinOS {
+		if ctx.Engine.Name == kubernetesEngine && ctx.ExecutableName == "" {
+			return errors.New("executable name is required for darwinOS on kubernetes engine (use --executable-name)")
```
andydotxyz, line 215: I don't understand why this is required. Is there not a sane default we can use?

Bluebugs: The problem we have is that k8s can not be assumed to have access on the host to the source code, but it is trying to emulate what fyne build does which require access to source code.

The simple solution at this point is to bite the bullet and move fully to fyne build. It is clearly a technical debt problem and we are adding more and more code around it instead of solving it.

Bluebugs:
This won't be necessary after merging #159 .

PR #159, opened 2022-12-30
Features/use fyne build

Bluebugs:
### Description:
This remove an historic piece of software that has done a lot to enable this tool, but it has past it's usefulness. By removing it and relying on fyne build directly, this fixes a lot of problem due to fyne-cross trying to reproduce fyne build and not always being able to.

I have tested all the platform impacted (darwin, linux, windows and freebsd), but it would be good to have more people testing this.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -141,16 +141,6 @@ func prepareIcon(ctx Context, image containerImage) error {
 		}
 	}
 
-	if image.OS() == "windows" {
-		// convert the png icon to ico format and store under the temp directory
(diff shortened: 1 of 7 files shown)


lucor:
As step forward to have fyne-cross as a subcommand of Fyne CLI this makes sense to have.

If we all agree, after merging this PR we'd release an alpha release of fyne-cross to help with testing, WDYT ?

Bluebugs:
> If we all agree, after merging this PR we'd release an alpha release of fyne-cross to help with testing, WDYT ?

That sounds like a very sensible approach to me.

lucor (approved):

PR #161, opened 2023-01-12
Address issue when using -dir and ctx.Icon can't be found.

Bluebugs:
### Description:
This make it possible to compile in a subdirectory using -dir by doing the proper lookup for the Icon specified on the command line.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/fyne-io/fyne-cross/internal/icon"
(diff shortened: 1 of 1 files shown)


internal/command/command.go
```diff
@@ -121,7 +121,12 @@ func cleanTargetDirs(ctx Context, image containerImage) error {
 // prepareIcon prepares the icon for packaging
 func prepareIcon(ctx Context, image containerImage) error {
 	if !ctx.NoProjectUpload {
-		if _, err := os.Stat(ctx.Icon); os.IsNotExist(err) {
+		iconPath := ctx.Icon
+		if ctx.Icon[0] != os.PathSeparator {
```
andydotxyz, line 125: This seems too simple - what about "./" or "../" - maybe it's need for another absolute check?

andydotxyz (changes requested):

internal/command/command.go
Bluebugs: In this case, we need to build an absolute path. I guess I could use filepath.IsAbs here.

andydotxyz: That is what I meant, sorry - my examples are bad. An absolute path is not defined as starting with the path separator on all OS.

andydotxyz (approved):

PR #163, opened 2023-01-14
Bugs/build sub package

Bluebugs:
### Description:
This restore past behavior being able to build a subdirectory.

It is to be noted that for this to be fully operational, new image with a new release of Fyne v2.3 is necessary.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.


Bluebugs:
android and ios do not support the -srcdir option. Seems that building a subdirectory for those target didn't work before either, maybe? I still have to look into it.

andydotxyz:
I don't understand how extracting binaries from the linux/BSD packages fixes the build location.

Bluebugs:
Hum, indeed, it is part of bringing back the feature of 1.3 that was storing the executable in a fyne-cross subdirectory. I should update the comment to be more precise.

internal/command/container.go
```diff
@@ -197,7 +197,7 @@ func fyneCommand(command string, ctx Context, image containerImage) ([]string, e
 		"-appVersion", ctx.AppVersion,
 	}
 
-	if ctx.Package != "." {
+	if ctx.Package != "." && image.OS() != androidOS {
```
andydotxyz: Should this note be a check for mobile rather than Android? iOS follows the same build constraints in this regard does it not?

Bluebugs: Actually not, as the iOS build are not going through the container build, but are straight host command (and are located in command.go).

andydotxyz: Might that ever change? The reason this path is needed is because of a problem with our go-mobile adopted code - not the Android specific builder...

Bluebugs: I don't think we have any plan or even idea on how we could do the ios build in the container at this point. But I realized that I actually refactored the code too much and that the container and the command file shared that code which was incorrect. Fixed now.

Bluebugs:
Got the build now working for all target with proper support for package and the proper filename.

andydotxyz (commented):
Seems sensible but it's a bit beyond me. Any chance @lucor could double check?

internal/command/command.go
```diff
@@ -319,7 +325,34 @@ func fyneReleaseHost(ctx Context, image containerImage) error {
...
+		return "", fmt.Errorf("could not package the Fyne app: %v", err)
 	}
-	return nil
+	return searchLocalResult(volume.JoinPathHost(workDir, "*"+ext))
+}
+
+func searchLocalResult(path string) (string, error) {
+	matches, err := filepath.Glob(path)
+	if err != nil {
+		return "", fmt.Errorf("could not find the ipa file: %v", err)
```
andydotxyz, line 336: I don't think this method is ipa specific

internal/command/command.go
```diff
@@ -319,7 +325,34 @@ func fyneReleaseHost(ctx Context, image containerImage) error {
...
+		}
+
+		if fi.ModTime().After(newestModTime) {
+			newest = match
+			newestModTime = fi.ModTime()
+		}
+	}
+
+	if newest == "" {
+		return "", fmt.Errorf("could not find the ipa file")
```
andydotxyz, line 355: as above - the extension isn't always spa

andydotxyz (changes requested):

internal/command/command.go
Bluebugs: Indeed, good catch.

andydotxyz (dismissed):

andydotxyz:
Can you please wait for @lucor to glance over it as well, this seems quite a big amount of changes to restore the behaviour?

andydotxyz:
Still looking good, what do you think @lucor ?

Also, should there be an issue created for the web sub-package build.

Bluebugs:
I need to try with a project that would compile with the web target and is in a sub package first. Terminal do not run in the web as we can't start a shell or anything like that.

lucor (approved):
Great work!
LGTM tested also against the `paw` project and it works as expected.
The only minor thing we noticed is `.js.map` not removed for the web build, but as per our chat discussion this is something not blocker and that will be addressed into a separated issue for fyne CLI

PR #175, opened 2023-02-24
Features/switch to 1.17

Bluebugs:
### Description:
After our discussion on slack, I have experimented switching go 1.17 metadata. This does properly build only the needed dependencies and work with go install. Binary size is as expected around 4MB as it was before this change and the build time is not changed. I would like to get everyone feedback on this bump go version bump.

### Checklist:
- [x] Tests included.
- [ ] Tests all pass.


README.md
```diff
@@ -42,14 +42,12 @@ Supported targets are:
 
 For go >= 1.16:
```
Jacalz, line 43: Shouldn't this be changed to Go 1.17 or reworded as it no longer makes sense to mention it?

go.mod
```diff
@@ -1,18 +1,59 @@
...
 require (
 	github.com/BurntSushi/toml v1.1.0
 	github.com/Kodeworks/golang-image-ico v0.0.0-20141118225523-73f0f4cfade9
-	github.com/andybalholm/brotli v1.0.4 // indirect
-	github.com/fyne-io/fyne-cross/internal/cloud v0.0.0
-	github.com/golang/snappy v0.0.4 // indirect
-	github.com/pierrec/lz4/v4 v4.1.14 // indirect
+	github.com/aws/aws-sdk-go v1.41.19
+	github.com/klauspost/compress v1.13.4
+	github.com/mholt/archiver/v3 v3.5.1
```
Jacalz, line 10: The bump to Go 1.17 should let you bump both of these versions to their latest versions, I think.

Jacalz (changes requested):
Nice work. Just a few notes online but will check this through locally once I have some more time on my hands :)

README.md
Bluebugs: So, the weird bit here is that we can still compile with go 1.14 (as the github action show). I do not really know what we should do/says here. I have not tested with 1.16.

go.mod
Bluebugs: Bumping those require code change I think. I can see if there is some minor bump possible that won't require code change.

Jacalz: The bump of compress should not need any code changes but it will not let us compile with older versions of Go though.

README.md
Jacalz: Ah, you're right. The new module syntax is compatible to run with older Go versions but developers with them locally will not be able to update/change the module. I guess we should decide on if we want to soft-support older Go or actually raise the version.

andydotxyz:
I'm happy if the others are :)

Jacalz (approved):

PR #177, opened 2023-03-16
Features/improve logs

Bluebugs:
### Description:
Silly enough, but all this log slow down significantly our download when using the k8s backend. It is more visible now with the MacOS SDK being provided as a mount point.

### Checklist:
- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):
Makes sense

PR #178, opened 2023-03-17
Make it possible to include the SDK in the darwin container.

Bluebugs:
### Description:
When running fyne-cross with k8s, it is more efficient to build an image that include the SDK. This patch make it possible for fyne-cross to assume the SDK was packaged in the container.

### Checklist:
- [x] Tests all pass.

### DIFF
--- a/internal/command/darwin.go
+++ b/internal/command/darwin.go
@@ -75,6 +75,8 @@ func (cmd *darwin) Parse(args []string) error {
 
 	flagSet.StringVar(&flags.MacOSXVersionMin, "macosx-version-min", "unset", "Specify the minimum version that the SDK you used to create the Darwin image support")
 
+	flagSet.BoolVar(&flags.ContainerIncludeMacOSXSDK, "container-include-macosx-sdk", false, "If set the macOS SDK is expected to be present in the container")
+
 	flagAppID := flagSet.Lookup("app-id")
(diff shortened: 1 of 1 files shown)


lucor (commented):
Makes sense to me.
What about adding a special value for the existing flag `sdk-path` like `bundled` or `included`. In theory we could also by default check for a bundled SDK before return an error. What do you think?

andydotxyz:
> In theory we could also by default check for a bundled SDK before return an error. What do you think?

I agree with this - even the new flag docs imply that it is an indication rather than a specified location. So why not just check if it exists before mounting an external location and/or erroring because one was not specified?

Bluebugs:
I am not to sure how you would check the existence of a file inside the container, any suggestion?

andydotxyz:
> I am not to sure how you would check the existence of a file inside the container, any suggestion?

I don't have a specific suggestion - it was just an idea. 
If it cannot be done then @lucor suggestion seems like a great one.

Bluebugs:
> `docker run ... fyneio/fyne-cross-images:darwin sh -c 'ls /sdk/path/to/file'` and check for the error

That's working, nice!

lucor (approved):
I'd add a comment to describe the objective of the ls command. Apart that looks good to me

Bluebugs:
Comment added!

andydotxyz (approved):
Nice

lucor (approved):
Great, thanks!

PR #179, opened 2023-03-17
zig does ignore ldflags passed by go, we need to specify them on zig …

Bluebugs:
### Description:
As zig ignore LDFLAGS, we need to provide them in an alternate way. Maybe it would be good to have `fyne build` detect zig and do that somehow, but that seems more complex work.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/command/windows.go
+++ b/internal/command/windows.go
@@ -213,16 +213,16 @@ func (cmd *windows) setupContainerImages(flags *windowsFlags, args []string) err
 		switch arch {
 		case ArchAmd64:
 			image.SetEnv("GOARCH", "amd64")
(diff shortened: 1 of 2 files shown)


andydotxyz:
Unfortunately I don't really understand this - the description says we are passing LDFLAGS in a different way, but the flags don't seem to match.
Is "subsystem-windows" the equivalent off specifying the windows build is a GUI app, or is something more clever happening?

Bluebugs:
Yes, it is the equivalent for zig.

andydotxyz (approved):

PR #180, opened 2023-03-24
Add ability to bypass darwin sdk on image test.

Bluebugs:
In the case of using Kubernetes backend, the container isn't started and accessible before we start the image with Prepare. It is simpler to just not try to test when we know what we are doing.

This was suggested during the previous and it is in practice a good solution.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/internal/command/darwin.go
+++ b/internal/command/darwin.go
@@ -67,7 +67,7 @@ func (cmd *darwin) Parse(args []string) error {
 	if runtime.GOOS == darwinOS {
 		flagSet.BoolVar(&cmd.localBuild, "local", true, "If set uses the fyne CLI tool installed on the host in place of the docker images")
 	} else {
-		flagSet.StringVar(&flags.MacOSXSDKPath, "macosx-sdk-path", "unset", "Path to macOS SDK [required]")
+		flagSet.StringVar(&flags.MacOSXSDKPath, "macosx-sdk-path", "unset", "Path to macOS SDK (setting it to 'ignore' will skip the in container sdk check) [required]")
 	}
 
 	// flags used only in release mode
--- a/internal/command/darwin.go
+++ b/internal/command/darwin.go
@@ -244,7 +244,7 @@ func (cmd *darwin) setupContainerImages(flags *darwinFlags, args []string) error
 				if err != nil {
 					return errors.New("macOSX SDK path is mandatory")
 				}
-			} else {
+			} else if flags.MacOSXSDKPath != "ignore" {
 				if _, err := os.Stat(flags.MacOSXSDKPath); os.IsNotExist(err) {
 					return errors.New("macOSX SDK path does not exists")
 				}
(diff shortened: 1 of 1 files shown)


lucor (dismissed):
should we document this value in the command flag help?

Bluebugs:
> should we document this value in the command flag help?

Indeed, good idea. Just done.

internal/command/darwin.go
```diff
@@ -67,7 +67,7 @@ func (cmd *darwin) Parse(args []string) error {
 	if runtime.GOOS == darwinOS {
 		flagSet.BoolVar(&cmd.localBuild, "local", true, "If set uses the fyne CLI tool installed on the host in place of the docker images")
 	} else {
-		flagSet.StringVar(&flags.MacOSXSDKPath, "macosx-sdk-path", "unset", "Path to macOS SDK [required]")
+		flagSet.StringVar(&flags.MacOSXSDKPath, "macosx-sdk-path", "unset", "Path to macOS SDK (setting it to 'ignore' will skip the in container sdk check) [required]")
```
andydotxyz, line 70: I can't quite follow "'ignore' will skip the in container sdk check" - I wonder if "ignore" is more about the implementation rather than the intent.
Would it not mean more to the developer if it was `(set to 'bundled' if the SDK is included in your image)`.
The original reads as though it is possible to have no SDK (i.e. ignore the error of none present)

andydotxyz (changes requested):
Can we make the developer facing parameters closer to the intent rather than the internal details required to make them work?

andydotxyz:
I may have misunderstood, but it seems like ignore isn't quite right - we are not supporting building macOS without an SDK are we?

internal/command/darwin.go
Bluebugs: `bundled` would work indeed. Will update.

internal/command/darwin.go
```diff
@@ -67,7 +67,7 @@ func (cmd *darwin) Parse(args []string) error {
 	if runtime.GOOS == darwinOS {
 		flagSet.BoolVar(&cmd.localBuild, "local", true, "If set uses the fyne CLI tool installed on the host in place of the docker images")
 	} else {
-		flagSet.StringVar(&flags.MacOSXSDKPath, "macosx-sdk-path", "unset", "Path to macOS SDK [required]")
+		flagSet.StringVar(&flags.MacOSXSDKPath, "macosx-sdk-path", "unset", "Path to macOS SDK (setting it to 'bundled' indicate that the sdk is expected to be in the container) [required]")
```
andydotxyz, line 70: "indicates"

andydotxyz (changes requested):
Typo?

lucor (approved):

andydotxyz (approved):
Looks good

PR #182, opened 2023-03-27
Remove gles tag now that we rely on fyne to do the right things for us.

Bluebugs:
### Description:
As highlighted during the review of PR 1.4.0, we do not need anymore the gles tag for linux arm target. I verified on my PineBook Pro and this work.

### Checklist:
- [x] Tests included.
- [x] Tests all pass.


lucor (approved):
Good catch. 
This probably is one of the thing left behind and no more needed since we have moved to fyne CLI for the build...

PR #186, opened 2023-04-14
Add ability to disable network in the container while building.

Bluebugs:
### Description:
This can only be used with vendored directory. It does provide two main benefit. First not setting up network, means that the build will not be able to get any data from the outside. It give you that guarantee that only the source code that is in your vendor directory was used. Second it is faster to setup a container without network.

### Checklist:
- [x] Tests all pass.

#### Where applicable:


Bluebugs:
Inspiration from https://www.redhat.com/sysadmin/speed-containers-podman-raspberry-pi .

Bluebugs:
@Jacalz enjoy :-)

lucor (approved):

Jacalz (approved):
Nice 👍

PR #193, opened 2023-05-04
Features/propagate goflags env

Bluebugs:
### Description:
This PR will pick GOFLAGS from fyne-cross host environment and propagate them to the call in the container.

#### Where applicable:

### DIFF
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -79,7 +79,7 @@ func AppendEnv(args []string, environs map[string]string, quoteNeeded bool) []st
 		if quoteNeeded && strings.Contains(v, "=") {
 			// engine requires to double quote the env var when value contains
 			// the `=` char
-			env = fmt.Sprintf("%q", env)
+			env = fmt.Sprintf("%s=%q", k, v)
 		}
 		args = append(args, "-e", env)
 	}
--- a/internal/command/docker.go
+++ b/internal/command/docker.go
@@ -77,7 +77,7 @@ func AppendEnv(args []string, environs map[string]string, quoteNeeded bool) []st
 	for k, v := range environs {
 		env := k + "=" + v
 		if quoteNeeded && strings.Contains(v, "=") {
-			// engine requires to double quote the env var when value contains
+			// engine requires to double quote the value when it contains
 			// the `=` char
 			env = fmt.Sprintf("%s=%q", k, v)
 		}
--- a/internal/command/docker_test.go
+++ b/internal/command/docker_test.go
@@ -268,6 +268,109 @@ func TestCmdEnginePodman(t *testing.T) {
 	}
 }
 
+func TestAppendEnv(t *testing.T) {
+	type args struct {
+		args        []string
+		env         map[string]string
+		quoteNeeded bool
+	}
+	tests := []struct {
+		name      string
+		args      args
+		wantStart []string
+		wantEnd   [][2]string
+	}{
+		{
+			name: "empty",
+			args: args{
+				args:        []string{},
+				env:         map[string]string{},
+				quoteNeeded: true,
+			},
+			wantStart: []string{},
+			wantEnd:   [][2]string{},
+		},
+		{
+			name: "quote needed",
+			args: args{
+				args:        []string{},
+				env:         map[string]string{"VAR": "value"},
+				quoteNeeded: true,
+			},
+			wantStart: []string{},
+			wantEnd:   [][2]string{{"-e", "VAR=value"}},
+		},
+		{
+			name: "quote not needed",
+			args: args{
+				args:        []string{},
+				env:         map[string]string{"VAR": "value"},
+				quoteNeeded: false,
+			},
+			wantStart: []string{},
+			wantEnd:   [][2]string{{"-e", "VAR=value"}},
+		},
+		{
+			name: "multiple",
+			args: args{
+				args:        []string{},
+				env:         map[string]string{"VAR": "value", "VAR2": "value2"},
+				quoteNeeded: true,
+			},
+			wantStart: []string{},
+			wantEnd:   [][2]string{{"-e", "VAR=value"}, {"-e", "VAR2=value2"}},
+		},
+		{
+			name: "multiple with args",
+			args: args{
+				args:        []string{"arg1", "arg2"},
+				env:         map[string]string{"VAR": "value", "VAR2": "value2"},
+				quoteNeeded: true,
+			},
+			wantStart: []string{"arg1", "arg2"},
+			wantEnd:   [][2]string{{"-e", "VAR=value"}, {"-e", "VAR2=value2"}},
+		},
+		{
+			name: "multiple with args and equal sign require quoting values",
+			args: args{
+				args:        []string{"arg1", "arg2"},
+				env:         map[string]string{"VAR": "value", "VAR2": "value2=2"},
+				quoteNeeded: true,
+			},
+			wantStart: []string{"arg1", "arg2"},
+			wantEnd:   [][2]string{{"-e", "VAR=value"}, {"-e", "VAR2=\"value2=2\""}},
+		},
+		{
+			name: "multiple with args and equal sign do not require quoting values",
+			args: args{
+				args:        []string{"arg1", "arg2"},
+				env:         map[string]string{"VAR": "value", "VAR2": "value2=2"},
+				quoteNeeded: false,
+			},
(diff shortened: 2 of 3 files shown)


Jacalz (changes requested):
Nice. Just a little comment in-line about not calling `os.Getenv()` twice to avoid needing two system calls if the environment variable is set up.

internal/command/docker.go
```diff
@@ -79,7 +79,7 @@ func AppendEnv(args []string, environs map[string]string, quoteNeeded bool) []st
 		if quoteNeeded && strings.Contains(v, "=") {
 			// engine requires to double quote the env var when value contains
 			// the `=` char
-			env = fmt.Sprintf("%q", env)
+			env = fmt.Sprintf("%s=%q", k, v)
```
andydotxyz, line 82: I wonder if this could have a unit test? It potentially could have a big impact on other variable appending and we have seen a few bugs recently that suggest that not all envs encode the same...

Bluebugs: Not to sure what the test could be that doesn't look like just enforcing the same code as here, since it is following docker flags interface. Maybe we should exec docker and podman in our tests for that?

Jacalz (dismissed):

internal/command/docker.go
andydotxyz: It could be factored out to a helper that formats the arts and demonstrates that non-trivial inputs (spaces etc) create the right format of string.
Just a little sanity checking really as a simple "x=y" seems like a potential spot for glitches in env and cmd handling.

Bluebugs: Such a tests would enforce and lock implementation details, that are not necessarily correct. That's where I have trouble imagining a test without having to spin a docker and podman command line that is not locking implementation, but actually enforce that it work.

andydotxyz: I am totally lost - why does passing an env into a printf require a docker container?

Bluebugs: > > The printf is just trying to mimics docker/podman expectation.
> 
> Well, what is the expectation? It must be documented somewhere. Without tests we are constructing parameters with printf and hoping they are right. With tests we can document the expected behaviour - then if something is wrong we will see it in the expectations wrong and we can fix that.

We are talking exactly about the same thing, but where I see a problem you seems to a solution :-) Writing such a test enshrine a belief in my opinion, or as you say it document our expectation (there is no specification of what docker parameters are, there is documentation that are open to interpretation as with all integration problem). This doesn't mean that passing this unit test is correct (or not passing it is incorrect). It doesn't give us any confidence that for our next release things are any better. Also if we are wrong about our assumption, the test reduce velocity in fixing the issue as we now have two points to fix.

To me, this would be using the wrong tool, unit test, for the job as this is an integration job. If we want confidence in the result and no slow down in the future that the only test that make sense to write. So back to my point, the test that have value in my opinion for this problem require to bring docker and podman (because they do not necessarily perfectly match each other and their behavior can change over time) as a dependency during testing and they will have to download some small distribution so we can execute `export`. Is that ok to bring in? I am guessing I will have to protect those test behind a tag and only run it during ci on github. I think podman only work on mac node, but not sure will have to check what github action can run.

andydotxyz: I don't understand the argument here - why can we not test that a string is correctly escaped given certain inputs?

There is a comment on the previous line stating "_engine requires to double quote the env var when value contains the `=` char_" - so there is an expectation set somewhere. Here the behaviour has changed but the comment has not - so which is wrong? I don't think it is easy to tell without an illustration such as in a unit test.

Bluebugs: > I don't understand the argument here - why can we not test that a string is correctly escaped given certain inputs?

We can always write code, but the first question is `should we?`. In this case, writing a string comparison as unit test for an integration problem is not going to get any confidence toward the fact that we have a working code, instead it enforce a belief, which can be proven wrong at any point. This type of code is clear technical debt. Test can also be a carrier of technical debt.

> There is a comment on the previous line stating "_engine requires to double quote the env var when value contains the `=` char_" - so there is an expectation set somewhere. Here the behaviour has changed but the comment has not - so which is wrong? I don't think it is easy to tell without an illustration such as in a unit test.

The comment was actually correct, the code wasn't. If you look at the code, we are quoting the environment now, while before the value was being quoted.

andydotxyz: If that was the case then the comment "double quote the env var when value contains the `=` char" would surely have read "double quote the env value when value contains the `=` char", as the environment variable is in the KEY=VALUE format is it not?

Jacalz: I agree with Andrew. It is better to verify that the code does what we think is right than not testing it at all.

Bluebugs: > If that was the case then the comment "double quote the env var when value contains the `=` char" would surely have read "double quote the env value when value contains the `=` char", as the environment variable is in the KEY=VALUE format is it not?

I see and understand that it might be confusing. In that case, any suggestion that would be less confusing for you?

Bluebugs: > It is better to verify that the code does what we think is right than not testing it at all.

I disagree on this statement. Tech debt for tests is worse than no tests. Never the less, it seems their is a consensus for having this kind of tests in, so I will oblige and add one.

Bluebugs: Test and comment change are in.

mikaelnousiainen:
Would be great to have this merged and released asap now that the tests are in!

internal/command/docker_test.go
```diff
@@ -273,6 +273,109 @@ func TestCmdEnginePodman(t *testing.T) {
...
+			wantEnd:   [][2]string{{"-e", "VAR=value"}, {"-e", "VAR2=value2=2"}},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := AppendEnv(tt.args.args, tt.args.env, tt.args.quoteNeeded)
+			for i, v := range tt.wantStart {
+				assert.Equal(t, v, got[i])
+			}
+			for i := len(tt.wantStart); i < len(got); i += 2 {
```
andydotxyz, line 365: isn't `len(tt.wantStart)` always going to be the same as `i`?
It took a while to realise this was contiguous, so I would suggest that I spans the two for loops so that is clear.

I also wondered, wouldn't it make sense to simplfy to just have all the output in a single `want` that is iterated?

andydotxyz (commented):
Personally I would have factored out the escaping code to a smaller scope so that could be tested with much simpler unit code.
However this certainly covers the usage - but as commented inline isn't it easier / clearer just to iterate over one `want` for the args list?

internal/command/docker_test.go
Bluebugs: The issue is that you can't rely on hash map order, which is why there is two loops. One for the part that should be stable and one for the part that can vary.

ofman (approved):

internal/command/docker_test.go
andydotxyz: Oh I see. But that wouldn't impact the usage of i - a single variable would better indicate that it is checking every item in the slices, even though split over two loops?

Bluebugs: I am not sure what you exactly mean. I reused the same name (`i`), but did want to use the cleaner, more readable, call to `range` when possible.

Bluebugs: This actually make the code a bit more complex to read. `range` does exit the loop with i == len(tt.wantStart) - 1, but if there is nothing in tt.wantStart, it exit with i == 0. So before I can use it I have to add a test to see if len(tt.wantStart) != 0, to add 1 back. This is a lot less readable at the end that with the current code I think. Do you have any suggestion on how to solve that? I can only imagine with going with the C style loop syntax instead of `range`.

andydotxyz: I think you could increment `i` inside the range loop too, if that's better?

Bluebugs: Do you mean that i would not be managed by range, but manually? Will have to try that.

Bluebugs: Not a big fan, but if that's what you had in mind, I am fine with it.

andydotxyz (approved):
Thanks for this

internal/command/docker_test.go
andydotxyz: The notation may not be totally awesome, but it helps to show we are actually only iterating once, despite two loops.
Thanks for the update.

PR #196, opened 2023-05-11
Update docs with note about Windows arm64 support

Jacalz:
### Description:

Windows arm64 has been available a while now but it seems like we missed adding it to the README.md file. This PR targets the `master` branch to make sure that users see it being supported.

On a side not, it would also be cool if [Geoffrey](https://fynelabs.com/geoffrey/) exposed support for it but it is obviously outside of the scope of this PR.

### Checklist:

- [ ] Tests included.
  - I did at least make sure to do a local build to double check that it was working :) 
- [x] Tests all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Supported targets are:
   -  linux/arm
   -  linux/arm64
   -  windows/amd64
+  -  windows/arm64
   -  windows/386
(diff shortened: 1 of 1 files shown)


Bluebugs (approved):
Sorry I missed this pr.

PR #206, opened 2023-09-02
Fix Android release that generates a .aab file

metal3d:
### Description:

Creating Android release generates a .aab file. The Build() function was working with *.apk files only. So the AAB file was not copied and triggered a SELinux error. The copy sh command also failed.

This commit fixes the problem by changing the pattern to match on copy, and the package name using ".aab" suffix.

NOTE: it could be better to use filepath.Glob() to find files and use os.Rename() instead.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (dismissed):
Oh thanks for catching this

andydotxyz:
The failing tests indicate we'll need to update the tested versions to match Fyne...

Bluebugs:
Yes, this require an image release with latest Fyne.

Bluebugs:
Do you mind synchronizing this with latest fyne-cross develop? I think it should be good to merge after that.

andydotxyz:
I have attempted a rebase myself to get this included

andydotxyz (approved):
All up to date and looking good, thanks

PR #225, opened 2024-03-16
Trigger CI before release of getting develop to master

Bluebugs:
### Description:
This will be next release of fyne-cross.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.


andydotxyz:
Sorry @Bluebugs can you be clear what the review is for here? The title just says "trigger CI before release" so I wanted to check...

Bluebugs:
The last pr had their tests failed because it wasn't merged with develop before getting in. I could have just merged develop with master, but that wouldn't have triggered any ci. This way, we get a pr that run the ci and people can review at the same time what is coming for fyne-cross next release.

Jacalz:
It looks from the commit list like fixuid was removed but it seems to still be in https://github.com/fyne-io/fyne-cross-images. Should it be removed from there as well? We should also make sure to get the latest Go 1.21 release in (1.22 is still a bit too young, 1.22.1 contained only security fixes).

Jacalz:
Opened https://github.com/fyne-io/fyne-cross-images/pull/43.

Bluebugs:
Image releases are independent from fyne cross release which kind of make me think we can't really remove fixuid from the image even if we don't use it as that will break previous user (break backward compatibility). This is something that likely require a major bump in version for the image and not necessary at the moment. The rest are good to do, but unrelated to this release.

andydotxyz (approved):
It looks good to me, but I'm no expert here

PR #243, opened 2024-04-28
Enable-macos-ci

Bluebugs:
### Description:
Turn back on CI for MacOS and Web target.

### Checklist:
- [ ] Tests included.
- [ ] Tests all pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - master
+      - develop
   pull_request:
   workflow_dispatch:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,10 @@
 name: "CI"
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,5 +1,10 @@
 name: "CI"
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
```
andydotxyz, line 5: Should the pushes to develop not be included?

andydotxyz:
@Bluebugs if you were able to rebase this on current master and add that it executes on develop as well then I think we should be in good shape.

.github/workflows/ci.yml
Bluebugs: Indeed

Bluebugs:
I don't know how it did happen, but the CI is now failing with Go 1.19 due to a dependencies that also break main/develop branch.

andydotxyz:
Yes I am not sure how it happened either. Some combination of deprecation and linting perhaps?
Do we need a smaller PR to resolve that first?

andydotxyz (approved):
Sweet thanks

PR #248, opened 2024-05-06
fix: update install suggestion for missing `fyne` command

nobe4:
### Description:

This makes the install process more inline with the CI workflow, where both are tested.

### Checklist:

- [ ] Tests included. N/A: no test checked this string, so I didn't add one.
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,11 +131,9 @@ jobs:
         working-directory: fyne-cross
         run: go install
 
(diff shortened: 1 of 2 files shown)


andydotxyz (changes requested):
Thanks for catching this. We don't need the "or go get" any more as the older way of installing binaries belongs to Go versions we don't support any more

andydotxyz (approved):
Great, thanks for this

Issue #250, opened 2024-05-14
`-cache` doesn't work on windows path from WSL2


williambrode:
### Describe the bug:
Running fyne-cross in WSL2 with the `-cache` option pointing to a windows path didn't fill in the cache.

### To Reproduce:
Steps to reproduce the behaviour:
1. In WSL2 install fyne-cross v1.4.0
2. Run `fyne-cross linux -cache /c/<path> ...`
3. Notice the path will be created but no files will be populated.
4. Doing this with a linux path (like ~/cache) works fine.

### Device and debug info (please complete the following information):

<summary>Device info</summary>

 - **OS:** WSL2 in windows 10
 - **Go version:** 1.22.3
 - **fyne-cross version:** 1.4.0
 - **Fyne version:** 2.4.4

Jacalz:
This is getting into border-line territory of potentially being a problem with WSL2. If this works for Linux paths but not emulated Windows paths, it looks like maybe the sandbox isn’t working? I’m not sure if even this is something we can do anything about.

Why not run fyne-cross on Windows directly?

williambrode:
After running into the issue again in CI I think I now understand that the problem was actually that fync-cross will appear to work with a relative path for `-volume` when it doesn't.  That is to say - it will create the directory, then when it tries to use the directory in docker it won't work because it doesn't pass the absolute path.

PR #252, opened 2024-05-14
[#250] Make -cache work with relative path.

williambrode:
### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


williambrode:
Perhaps a maintainer could suggest where to add a test for this?  I could add it around the immediate function (Mount) but then I'd just be testing the specific implementation.

Bluebugs:
We currently have mostly unit test in the form of things that test how a function will provide a result depending on its input. That would be something you could easily add as a new file in `internal/volume`.

But these are not super useful as it only prove that things match your expectation, not that it work.

We do have some better/more useful end 2 end kind of tests with the CI that build full application using fyne-cross, but they are too broad and we don't get any code coverage out of them. I would love to see some more integration tests that actually show fyne-cross work and provide proper tests coverage. I think that is really further away at the moment.

andydotxyz (commented):
Looks good, though it may imply the same should be done to the work dir in case it too is relative?

Regarding testing it may be that moving to helper methods which are easily tested is the way - a method named "Mount" would not necessarily be expected to do so much logic as it sounds pretty clear that it's mounting the paths. If the "cacheDirAbsOrDefault" helper existed then mount would essentially set up the object and create the dirs, a cleaner function overall I'd think.

andydotxyz:
Tidying up and figured this should land, we can adjust tests and Go versions etc as we migrate to tools

PR #254, opened 2024-05-15
Performance improvement by avoiding running commands in docker

williambrode:
I noticed that fyne-cross running on my gitlab CI was taking a very long time to clean directories that don't even exist.  I found that because these cleans are run using docker - it takes a while just to spin up the container to run them.

 

### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -97,20 +97,18 @@ Use "fyne-cross <command> -help" for more information about a command.
 func cleanTargetDirs(ctx Context, image containerImage) error {
 
 	dirs := map[string]string{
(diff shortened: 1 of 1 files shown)


andydotxyz:
As noted on our chat channels, I am not sure if it was important that the old way was used - hopefully @Bluebugs, @lucor or @Jacalz can comment?

williambrode:
This is a huge performance improvement - I think it may be worth trying unless we know a reason not to do it.

Jacalz:
I think we've all been rather busy unfortunately. I'll try to have a look first things after my vacation.

Bluebugs (approved):

PR #271, opened 2024-10-23
Add support for extra mount

DanielRiera:
### Description:
Add "extra-mount" params for add volumes to docker


andydotxyz (changes requested):
I'm no expert at this unfortunately - but I think that adding such functionality would need to be reflected in the podman engine code as well.

andydotxyz:
Also, merging in from the upstream branch should fix the test failure

andydotxyz:
We can't land without the suggested changes, it's been a long time so closing

PR #334, opened 2025-09-29
Use correct flags after tools change

sdassow:
### Description:

Sync changed Fyne CLI tool flags after report on Slack.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.


andydotxyz:
Thanks for this. I guess it will require a new image to support this as well,  because the tool installed on them is still the deprecated one I think?

sdassow:
> Thanks for this. I guess it will require a new image to support this as well, because the tool installed on them is still the deprecated one I think?

That's what I've been assuming as well so far and haven't found out how to do yet.

Bluebugs:
You may want to detect the tool version in the image to avoid breaking user experience.

andydotxyz:
> You may want to detect the tool version in the image to avoid breaking user experience.

That's a good point - switching between this new parameter set vs older one based on the result of `fyne version`.

It would be pretty painful otherwise - the alternative being forcing the container update on this update - but doing that we would have to have the images out first, which could cause breakage for anyone that updates the images before this update is released...

internal/command/command.go
```diff
@@ -160,26 +161,46 @@ func checkFyneBinHost(ctx Context) (string, error) {
...
+	return ""
+}
+
 func fyneCommand(binary, command, icon string, ctx Context, image containerImage) []string {
 	target := image.Target()
+	fyneVersion := fyneCommandVersion(binary)
+
+	appBuildOpt := "-app-build"
+	appVersionOpt := "-app-version"
+	if semver.Compare(fyneVersion, "v1.6.0") < 0 {
```
andydotxyz: Is this correct? The tool version before this release was v2.6.x I think and we went to lower version with the new location.
I could be misreading this but I think really we need something like is it less than v2.0.0 - assuming we remove the migration code before the tool actually does have a breaking 2.0 release...

sdassow: Yeah, just realized as well and changing it.

sdassow:
If I understand everything, then it's currently broken in the fyne-cross image, and this PR should fix it. And for that to take effect we need to build a new image with these changes first, right?
Other than that, the linter fails because of an old AWS SDK version, so the merge needs some force I think.

andydotxyz:
> If I understand everything, then it's currently broken in the fyne-cross image, and this PR should fix it. And for that to take effect we need to build a new image with these changes first, right? Other than that, the linter fails because of an old AWS SDK version, so the merge needs some force I think.

I don't think that is the case - Fyne-cross is not included in the image (fyne is).

I'm not sure why it errors with these changes - perhaps it is not installing in the action correctly, or maybe the fix still has problems?

internal/command/container.go
```diff
@@ -239,16 +239,16 @@ func fyneRelease(ctx Context, image containerImage) error {
 	case androidOS:
 		workDir = volume.JoinPathContainer(workDir, ctx.Package)
 		if ctx.Keystore != "" {
-			args = append(args, "-keyStore", ctx.Keystore)
```
andydotxyz, line 242: These also need a version check.

andydotxyz:
The macOS issue may be fixed by updating the runner, it has a v1 release now: 
`douglascamata/setup-docker-macos-action@v1.0.0`

andydotxyz:
Oh I figured out the problem @sdassow - the `fyne` binary being checked is the one on the /host/ computer.
That is correct for Darwin but for all others you need to execute `fyne version` in the container!

sdassow:
> Oh I figured out the problem @sdassow - the `fyne` binary being checked is the one on the /host/ computer. That is correct for Darwin but for all others you need to execute `fyne version` in the container!

Oh! Haven't looked at it from that perspective, thanks for the pointer, checking that next.

internal/command/container.go
sdassow: Fixed, and improved the version checks on the way.

sdassow:
> Oh I figured out the problem @sdassow - the `fyne` binary being checked is the one on the /host/ computer. That is correct for Darwin but for all others you need to execute `fyne version` in the container!

But fyne-cross detects the version now, so it should work either way, that's what is throwing me still off. The version check uses the same binary path, so the check should kick in. The only way this doesn't happen would be if the fyne-cross code used is the old one, hence my suspicion that it's in some image, and that's also how I still understand the CI logs after reading them a few times more.

andydotxyz:
I don't quite understand that comment sorry. Fyne cross sits on the host and prepares a container which contains `fyne` and all the Go tools. The execution of `fyne` (except for Darwin) will happen inside the container, but the code you have included to ask what version is running executes on the host.

If, for example, you are running on Linux or BSD you could try removing the `fyne` binary on the current computer then run fyne-cross. In this case your version check will fail but the build would otherwise complete.

Look at usages of the `fyneCommandContainer` command inside internal/command/. You can also see how the `fynePackageHost` vs `fynePackage` functions differ based on which executable should be used to bundle.

andydotxyz:
I think there is a misunderstanding somewhere - the Fyne version was already printed in debug mode.

Here is the output locally when I print the host versions and then execute fyne cross. You can see the docker image contains an older Fyne CLI and it is printed by previously released fyne-cross version:

```
andy@nostromo ~/C/Fyne/fyne  (develop)
><> fyne version                                                                                                    [21:30:51]
fyne cli version: v1.6.3-0.20251005194804-555a71945e3b
fyne library version: (not found)
andy@nostromo ~/C/Fyne/fyne  (develop)
><> fyne-cross android --debug -app-id io.fyne.hello cmd/hello                                                      [21:30:52]
[i] Target: android/multiple
&command.localContainerImage{baseContainerImage:command.baseContainerImage{arch:"multiple", os:"android", id:"android", env:map[string]string{}, tags:[]string(nil), mount:[]command.containerMountPoint{command.containerMountPoint{name:"project", localHost:"/home/andy/Code/Fyne/fyne", inContainer:"/app"}, command.containerMountPoint{name:"cache", localHost:"/home/andy/.cache/fyne-cross", inContainer:"/go"}}, DockerImage:"docker.io/fyneio/fyne-cross-images:android"}, runner:(*command.localContainerEngine)(0xc0000dc090)}
[✓] "temp" dir cleaned: /home/andy/Code/Fyne/fyne/fyne-cross/tmp/android
[✓] "bin" dir cleaned: /home/andy/Code/Fyne/fyne/fyne-cross/bin/android
[✓] "dist" dir cleaned: /home/andy/Code/Fyne/fyne/fyne-cross/dist/android
[i] Checking for go.mod: /home/andy/Code/Fyne/fyne/go.mod
[✓] go.mod found
[i] Packaging app...
/usr/bin/podman run --rm -t -w /app -v /home/andy/Code/Fyne/fyne:/app:z -v /home/andy/.cache/fyne-cross:/go:z --userns keep-id -e use_podman=1 --arch=amd64 -e CGO_ENABLED=1 -e GOCACHE=/go/go-build docker.io/fyneio/fyne-cross-images:android /usr/local/b
in/fyne version
fyne cli version: v2.5.5
/usr/bin/podman run --rm -t -w /app/cmd/hello -v /home/andy/Code/Fyne/fyne:/app:z -v /home/andy/.cache/fyne-cross:/go:z --userns keep-id -e use_podman=1 --arch=amd64 -e CGO_ENABLED=1 -e GOCACHE=/go/go-build docker.io/fyneio/fyne-cross-images:android /usr/local/bin/fyne package -os android -name fyne -icon /app/fyne-cross/tmp/android/Icon.png -appBuild 1 -appVersion 1.0.0 -appID io.fyne.hello -release
```

andydotxyz:
Looks like there are 2 failures, the lint which we will have to ignore for now and the other is:

"The configuration 'macos-13-us-default' is not supported"
(I guess GitHub supported platforms changed recently)

andydotxyz:
Well done on fixing the CI config.
I guess the remaining issue is related to the macOS build running the final "fyne package" on macOS directly instead of in the container (for XCode access).

andydotxyz:
If you're ignoring version macOS then I assume the expected setup is the tools version (not deprecated Fyne one). In which case the Darwin GitHub runner will need to be updated to install that tool so the tests pass

andydotxyz (approved):
Tested on Linux and on macOS with the latest images, working great.
Thanks so much!

PR #339, opened 2026-01-14
Update fyne cli version

murrayinglis:
### Description:

Fixes #(issue)

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -213,9 +213,9 @@ jobs:
         working-directory: fyne-cross
         run: go install
 
-      - name: Install Fyne
+      - name: Install Fyne CLI tools
         run: |
-          go install fyne.io/fyne/v2/cmd/fyne@latest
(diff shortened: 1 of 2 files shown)


andydotxyz:
This is quite a brittle change - it requires that the fyne tool inside the image matches what is expected by fyne-cross. This coupling will break a lot of builds as images and fyne-cross can be updated separately.

I recommend you read up and contribute to https://github.com/fyne-io/fyne-cross/pull/334 instead.

murrayinglis:
> This is quite a brittle change - it requires that the fyne tool inside the image matches what is expected by fyne-cross. This coupling will break a lot of builds as images and fyne-cross can be updated separately.
> 
> I recommend you read up and contribute to #334 instead.

Sure, will contribute to that so we can have a version check for the CLI tools

andydotxyz:
Thanks!

PR #343, opened 2026-05-17
Adding iossimulator checks

### LINKED ISSUE #244: Add necessary secret to GitHub Actions to run iOS build
We currently do not make any attempt to build for iOS in our CI as we do not have the necessary secret setup for that purpose.

Without the secret set, it is not possible to turn on iOS CI.

Add the following secrets to this github repository:
- *P12_CERTIFICATE_BASE64*: base64 encoded value of the .p12 file (`base64 -i certificates.p12`)
- *P12_PASSWORD*: password for the .p12 certificate above
- *PROVISION_PROFILE_BASE64*: base64 encoded value of a .mobileprovision file (`base64 -i file.mobileprovision`)
- *KEYCHAIN_PASSWORD*: a random password that will be used for the temporary keychain during the build

andydotxyz:
It won't cover signatures but will check the build capability.

Fixes #244

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,13 +98,10 @@ jobs:
             args: -app-id calc.sha${{ github.sha }}
           # macos versions running in a VM on Mac M1 are incompatible with docker/podman
             host: macos-15-intel
+          - os: iossimulator
+            args: -app-id calc.sha${{ github.sha }}
+            host: macos-15-intel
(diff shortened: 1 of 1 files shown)


andydotxyz:
It was a nice idea, but fyne-cross does not support iossimulator target!

PR #355, opened 2026-05-26
Migrate CLI module to match Fyne CLI tools

sdassow:
### Description:

As a first step towards merging with fyne tools, only replace the CLI module, and should not change behaviour otherwise.

### Checklist:

- [ ] Tests included.
- [ ] Tests all pass.

#### Where applicable:


andydotxyz (changes requested):
This appears to be a breaking change - any parameters passed with the single-dash prefix will no longer work as it's moved to two. Can we support both to avoid this break?
[✗] invalid value "test" for flag -app-build: parse error

sdassow:
> This appears to be a breaking change - any parameters passed with the single-dash prefix will no longer work as it's moved to two. Can we support both to avoid this break? [✗] invalid value "test" for flag -app-build: parse error

Hm, are you sure? The `-app-build` in your example fails because it's expecting a number.
It works for me as expected with single and double dash options.

